PR #1, opened 2023-06-06
Switch to use node API defined in CLI

tmeasday:
This uses the node API defined in this PR: https://github.com/chromaui/chromatic-cli/pull/756, currently published as `chromatic@6.19.5`. It also uses the new (experimental) server channel API to communicate with the CLI.

<details>
  <summary>📦 Published PR as canary version: <code>0.0.2--canary.1.1f74baa.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.2--canary.1.1f74baa.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.2--canary.1.1f74baa.0
  ```
</details>


.storybook/local-preset.js
```diff
@@ -8,3 +11,44 @@ function managerEntries(entry = []) {
...
+  const app = express();
+
+  app.post("/build", async (req, res) => {
+    let sent = false;
+    res.setHeader("access-control-allow-origin", "*");
+
+    await run({
+      flags: {
+        projectToken: "00baf09dbbe8",
+      },
```
tmeasday, line 26: @ghengeveld it feels to me like we should pass this as an option, not a flag (we don't need the CLI to do any parsing, etc). However, currently the `getOptions` code will throw if there's no `projectToken` set in the flags. Let's hash this out.

.storybook/local-preset.js
```diff
@@ -8,3 +11,44 @@ function managerEntries(entry = []) {
...
+  // Temporarily create a server as a psuedo server channel. This is throwaway code
+  const app = express();
+
+  app.post("/build", async (req, res) => {
+    let sent = false;
+    res.setHeader("access-control-allow-origin", "*");
+
+    await run({
+      flags: {
+        projectToken: "00baf09dbbe8",
```
ndelangen, line 25: I imagine we'd want to ask the user to put this in their `main.ts` file?

maybe:
```ts
const { projectToken } = await options.presets.apply('chromatic')
```

tmeasday: @ndelangen sure, but where do you get a reference to `options` from?

ndelangen: Any preset value can access the options object as a second argument

ndelangen: @tmeasday you can gain access like so:

Because you're not really interested in hooking into/modifying anything, you can use any presetValue you know storybook will call, like `stories`, `core`, `addons` and return the value unmodified.

```ts
export default {
  addons: async (current, options) => {
    // now you have access to options
    // const out = await options.presets.apply('whatever', 'default-value')
    
    // return unmodified
    return current;
  }
}
```

ndelangen (approved):

PR #9, opened 2023-06-26
Stricter linting and improved DX

ghengeveld:
This follows (some) advice from a [great talk](https://www.youtube.com/watch?v=sSJBeWPIipQ) on configuring ESLint for a TS React project.

Warnings are configured to break the build, so we don't get sloppy and ignore them.
Unused eslint-disable directives are reported so they behave similar to `@ts-expect-error`, which means we'll know when such a directive can be removed.
VSCode is configured with recommended extensions and sensible format/fix settings, and to highlight lint errors as warnings (yellow squigglies) to disambiguate from type errors.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,3 @@
+{
+  "recommendations": ["esbenp.prettier-vscode", "dbaeumer.vscode-eslint"],
+}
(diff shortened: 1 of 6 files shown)


tmeasday (approved):
Sounds good to me

PR #10, opened 2023-06-28
Show project picker when no project id is in `main.js`

weeksling:
https://github.com/chromaui/addon-visual-tests/assets/5882088/536c660f-17bc-4929-9445-e267804e4cf9

Adds a parameter for the projectId and lets the user select between projects they are associated with. Includes serverChannel method to mutate the storybook configuration when setting the projectId & projectToken.

Not working yet:
  - useParameters pulls values from preview.ts, but not from addon options. Need to find a way to use option parameters from main.ts instead.
 
Pull request base is test-screens to make the changes here easier to diff. Should be changed to main prior to merging.
This depends on related Chromatic api PR https://github.com/chromaui/chromatic/pull/7460

<details>
  <summary>📦 Published PR as canary version: <code>0.0.4-canary.10.f8c401c.0</code></summary>
  <br />
  
  ```bash
  npm install @chromaui/addon-visual-tests@0.0.4-canary.10.f8c401c.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.4-canary.10.f8c401c.0
  ```
</details>


src/utils/useProjectId.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const useProjectId = (): [
+  projectId: string,
+  updateProject: (projectId: string, projectToken?: string) => void,
+  projectIdChanged: boolean,
+  clearProjectIdChanged: () => void
+] => {
+  //
+  // Ideally this should only be configured once in main.ts. We'll need to figure out how to do that. This should at least read from that config for now.
+  const configuredProjectId = useParameter(PROJECT_PARAM_KEY, null);
```
weeksling, line 16: @ndelangen do you know of any way I could pull this configuration from the addon options in main.ts instead of preview.tsx (parameters)?

Our serverChannel methods seem to have access to options for projectToken, so I've left that value there. It seems like it might be better for our projectId to also be configured in `main.ts`.

.storybook/main.ts
```diff
@@ -17,6 +17,7 @@ const config: StorybookConfig = {
       name: "../dist/index.js",
       options: {
         projectToken: "00baf09dbbe8",
+        projectId: "6480e1b0042842f149cfd74c", // Default to the the production project of this addon
```
weeksling, line 20: @ndelangen related to the other question: this is where it seems like we would want to configure this projectId. I added this here as an example, it is not used. Currently, I don't know of any way to access these options inside of the addon Panel.tsx

tmeasday:
@ndelangen can you take a look, especially at the preset code-- @ghengeveld can you review the UI code.

I have it working doing the following:

1. Reading project id + token from `main.js`
2. Running a git command on startup
3. Setting all of the above in `process.env`
4. Writing to `main.js` when you select a project.

What's missing/broken right now is:

1. Because we use env to set initial values, if you select a project and refresh the browser, you'll lose track of the selected project. 
  - We could store the selected project in local storage to avoid this
  - Or we could change it to request the selected project + git info via an event on bootup as originally planned. WDYT @ndelangen?

2. Similarly if you change branch or commit but there's initial value built into the SB it'll go out of date if you refresh (although in that case I expect to poll for values). So maybe events is better?

3. The branch behaviour is weird in other ways but I'm not looking to fix that in this PR.

.storybook/main.ts
tmeasday: We sorted it out!

ndelangen:
> Or we could change it to request the selected project + git info via an event on bootup as originally planned. WDYT

I feel like that's what I've been saying for a while, that will yield the best result, no weird behavior for anyone.
It won't support static build, but the addon won't support static storybook anyway.

src/index.ts
```diff
@@ -1,8 +1,17 @@
...
+import { promisify } from "util";
 
-import { BUILD_STARTED, START_BUILD } from "./constants";
+import { BUILD_STARTED, START_BUILD, UPDATE_PROJECT, UpdateProjectPayload } from "./constants";
+import { findConfig } from "./utils/storybook.config.utils";
+
+const {
+  CHROMATIC_BASE_URL = "https://www.chromatic.com",
+  CHROMATIC_ADDON_NAME = "@chromaui/addon-visual-tests",
+} = process.env;
```
ndelangen, line 14: I recommend using globals over `process.env`.

But if you're going to use `process.env`, please use the full path, and don't destructure:
```
const CHROMATIC_BASE_URL = process.env.CHROMATIC_BASE_URL || 'https://chromatic.com';
```

This way you won't get the entire `process.env` stringified into your code (bloat).

src/index.ts
```diff
@@ -33,10 +46,58 @@ async function serverChannel(channel: Channel, { projectToken }: { projectToken:
...
+        }
+      );
+
+      MainConfig.setFieldValue(["addons"], updatedAddonsConfig);
+      await writeConfig(MainConfig);
+    }
+  );
+
   return channel;
 }
```
ndelangen, line 77: I'd expect a 
```
channel.on(
    REQUEST_PROJECT,
```
...here as well, that allows the front-end to request/receive the projectId

src/index.ts
```diff
@@ -33,10 +46,58 @@ async function serverChannel(channel: Channel, { projectToken }: { projectToken:
...
+  const commit = (await execPromise("git log -n 1 HEAD --format='%H'")).stdout.trim();
+  return { branch, commit };
+}
+
+const config = {
   managerEntries,
   experimental_serverChannel: serverChannel,
+  // FIXME: we should only do this in dev
+  env: async (env: Record<string, string>, { projectId }: { projectId: string }) => {
+    const { branch, commit } = await getGitInfo();
```
ndelangen, line 92: About the TODO:
The second argument is the `Options` object, off which you can extract the `configType` property.
That property will be either `development` or `production`.

src/utils/useProjectId.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const { CHROMATIC_PROJECT_ID } = process.env;
+
+export const useProjectId = (): [
+  projectId: string,
+  updateProject: (projectId: string, projectToken?: string) => void,
+  projectIdChanged: boolean,
+  clearProjectIdChanged: () => void
+] => {
+  const [projectId, setProjectId] = React.useState<string | null>(CHROMATIC_PROJECT_ID);
```
ndelangen, line 14: Instead of using the env variable here we want to emit an event, and "await" the response.
That way we always get the fresh value, not what it was when storybook started.

src/index.ts
tmeasday: This is `index.js` (a preset) it isn't replaced.

tmeasday: > I recommend using globals over process.env.

Do we have a way to do this in a builder agnostic way?

src/index.ts
tmeasday: Not currently implemented. We just read it on start up from env, and if/when the user picks a new project.

src/utils/useProjectId.ts
tmeasday: I commented about this in the PR description: https://github.com/chromaui/addon-visual-tests/pull/10#issuecomment-1633606915

Do you agree this is OK for now and we can iterate as we start polling the branch/commit info (and thus need some events to update the manager w/ new info)?

src/index.ts
ndelangen: Ow, whoops, I didn't catch this will be run in node when I was reviewing it.. I'll discard this recommendation!

src/index.ts
ndelangen: With the code above, it can't be super difficult to implement reading the value, right?

src/utils/useProjectId.ts
ndelangen: Yeah, it's OK with the caveat that you'll need to restart storybook after setting the projectId.
That's OK for internal use, I'd say.

src/index.ts
tmeasday: Why would we need to read the value? We already get it from the preset function.

tmeasday:
> I feel like that's what I've been saying for a while, that will yield the best result, no weird behavior for anyone.
> It won't support static build, but the addon won't support static storybook anyway.

I guess so. But it seems overly complex to have 2 events just to request some data. Is that the pattern we want to lay down for client initiated server->client communication? 

Other options:
  - HTTP requests rather than server channel (like `/index.json`) for data requests
  - Some API that abstracts away how it is implemented (which I think you also suggested).

My inclination is to take it a bit further like this just to gather information and then thrash it out very soon.

src/index.ts
```diff
@@ -33,10 +46,62 @@ async function serverChannel(channel: Channel, { projectToken }: { projectToken:
...
   return channel;
 }
 
-module.exports = {
+// TODO: use the chromatic CLI to get this info? Or should we make it core to SB?
+const execPromise = promisify(exec);
+async function getGitInfo() {
+  const branch = (await execPromise("git rev-parse --abbrev-ref HEAD")).stdout.trim();
+  const commit = (await execPromise("git log -n 1 HEAD --format='%H'")).stdout.trim();
+  return { branch, commit };
```
ghengeveld, line 84: I think we'll need `slug` as well, to properly handle forks.

src/index.ts
ndelangen: Because: the preset function will get a stale value. We don't re-read the files when calling the preset function.

src/Panel.tsx
```diff
@@ -33,16 +39,34 @@ export const Panel = ({ active }: PanelProps) => {
...
   if (!accessToken) return <Authentication key={PANEL_ID} setAccessToken={setAccessToken} />;
 
+  if (!projectId)
+    return (
+      <Provider key={PANEL_ID} value={client}>
+        <LinkProject onUpdateProject={updateProject} />
+      </Provider>
+    );
+
+  if (projectId && projectIdChanged) {
```
ghengeveld, line 58:
```suggestion
  if (projectIdChanged) {
```

src/index.ts
tmeasday: I also want to add the "uncommitted hash" or similar later also.

ghengeveld:
I added `slug` to the `lastBuild` field args. This requires [an update on the API](https://github.com/chromaui/chromatic/pull/7503) side.

ghengeveld (approved):

PR #11, opened 2023-07-12
Set sidebar status from build result

tmeasday:
Use the new `experimental_updateStatus` API to set the sidebar status as build results come in.

image

It's not totally clear what statuses we should be setting (`success` or `unknown` for passing/accepted tests?). There are a lot of details to work through, but the wiring is there.

---

Note to run this right now you need to install the addon into a sandbox on the [release/7.2 branch](https://github.com/storybookjs/storybook/tree/release/7.2), like so:

```
  addons: [
    '@storybook/addon-links',
    '@storybook/addon-essentials',
    '@storybook/addon-onboarding',
    '@storybook/addon-interactions',
    {
      name: '/Users/tom/GitHub/chromaui/addon-visual-tests/dist/index.js', // or similar
      options: {
        projectId: '64acdc2151cdda2cb8b7eaaf',
        projectToken: 'chpt_4c07aba6aed67ce',
      },
    },
  ],
```

<details>
  <summary>📦 Published PR as canary version: <code>0.0.6--canary.11.bd86cca.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.6--canary.11.bd86cca.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.6--canary.11.bd86cca.0
  ```
</details>


src/screens/VisualTests/VisualTests.tsx
```diff
@@ -193,14 +196,19 @@ export const VisualTests = ({
...
+
   useEffect(() => {
-    if (isRunning && data?.build && "result" in data.build) {
-      setIsOutdated(false);
-      setIsRunning(false);
+    if (!build) return;
+
+    if ("tests" in build) {
+      updateBuildStatus(build);
     }
```
ndelangen, line 206: Is this the reason you added the memoization just above?

A simple check here, would save complexity and improve performance.
Sure we can memoize and make the callback do nothing.. but what if we just didn't call the setState when we know it's not needed?

src/screens/VisualTests/VisualTests.tsx
```diff
@@ -157,6 +158,7 @@ interface VisualTestsProps {
   setAccessToken: (accessToken: string | null) => void;
   setIsOutdated: (isOutdated: boolean) => void;
   setIsRunning: (isRunning: boolean) => void;
+  updateBuildStatus: (build: StartedBuild | CompletedBuild) => void;
   storyId: string;
```
ndelangen, line 162: It seems these types are defined multiple times, and should likely be shared/extracted.

src/utils/testsToStatusUpdate.ts
```diff
@@ -0,0 +1,49 @@
...
+
+// Just the bits we need
+type ReducedTest = Pick<TestFieldsFragment, "story" | "status">;
+
+const statusOrder: StoryStatus[] = ["unknown", "pending", "success", "warn", "error"];
+function chooseWorseStatus(status: StoryStatus, oldStatus: StoryStatus | null) {
+  if (!oldStatus) return status;
+
+  return statusOrder[Math.max(statusOrder.indexOf(status), statusOrder.indexOf(oldStatus))];
+}
```
ndelangen, line 27: Hmm probably not great to re-implement this, But I don't really have a great place put this so i can export it in storybook..

src/utils/testsToStatusUpdate.ts
```diff
@@ -0,0 +1,49 @@
...
+  const update = Object.fromEntries(
+    Object.entries(storyIdToStatus).map(([storyId, status]) => [
+      storyId,
+      {
+        status,
+        title: "Visual Tests",
+        description: "Chromatic Visual Tests",
+      },
+    ])
+  );
```
ndelangen, line 46: We could make this get the existing status state, and only set when things are different.

src/screens/VisualTests/VisualTests.tsx
tmeasday: Either way we need to check if the build is different to last time, I think it's going to be faster to stringify the status update than the whole build. The alternative would maybe be to look at which tests have changed since the last time we considered updating, but I'm not sure if it's worth it without seeing perf issues.

tmeasday: > Is this the reason you added the memoization just above?

I added the memoization to avoid a set state loop -- each time the `VisualTests` component is rendered, it otherwise calls `setStatus` which changes the state of the manager, forcing a re-render of the `VisualTests` component.

src/utils/testsToStatusUpdate.ts
tmeasday: Yes, I had the same thought process...

src/utils/testsToStatusUpdate.ts
tmeasday: IMO SB should do this for me, don't you think?

src/utils/testsToStatusUpdate.test.ts
```diff
@@ -15,7 +15,7 @@ it("handles single test", () => {
     {
       "story--id": {
         "description": "Chromatic Visual Tests",
-        "status": "success",
```
ndelangen: That's not a valid status value. `I think you're looking for `unknown`:
https://github.com/storybookjs/storybook/blob/2e2f7dd191c94bd07b856264c81023223d8e8ecf/code/lib/types/src/modules/api-stories.ts#L176

src/screens/VisualTests/VisualTests.tsx
tmeasday: Do you have a suggestion on how to do that?

tmeasday:
I moved the memoization to the use effect, can you take a look @ndelangen?

src/screens/VisualTests/VisualTests.tsx
```diff
@@ -197,17 +197,31 @@ export const VisualTests = ({
...
+      buildStatusUpdate,
+      string: JSON.stringify(buildStatusUpdate),
+    };
+    if (buildStatusUpdate) {
+      updateBuildStatus(buildStatusUpdate);
     }
-  }, [setIsOutdated, setIsRunning, updateBuildStatus, isRunning, build]);
+    // We use the stringified version of buildStatusUpdate to do a deep diff
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [JSON.stringify(buildStatusUpdate), updateBuildStatus]);
```
ndelangen, line 224: that works!

ndelangen (approved):

src/utils/testsToStatusUpdate.ts
tmeasday: Let's put a pin in this idea until we get more informatoin.

PR #15, opened 2023-07-14
add code to read the projectId over event

ndelangen:
@tmeasday sort of a POC how I'd do it.

I did not run the code, so might not work fully

<details>
  <summary>📦 Published PR as canary version: <code>0.0.4-canary.15.4c0149a.0</code></summary>
  <br />
  
  ```bash
  npm install @chromaui/addon-visual-tests@0.0.4-canary.15.4c0149a.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.4-canary.15.4c0149a.0
  ```
</details>

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -10,6 +10,8 @@ export const START_BUILD = `${ADDON_ID}/startBuild`;
 export const BUILD_STARTED = `${ADDON_ID}/buildStarted`;
 
 export const UPDATE_PROJECT = `${ADDON_ID}/updateProject`;
+export const READ_PROJECT = `${ADDON_ID}/readProject`;
+export const RESPOND_PROJECT = `${ADDON_ID}/respondProject`;
 export type UpdateProjectPayload = {
   projectId: string;
   projectToken: string;
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,11 +1,19 @@
 import type { Channel } from "@storybook/channels";
 import { readConfig, writeConfig } from "@storybook/csf-tools";
+import { StorybookConfig } from "@storybook/types";
 import { exec } from "child_process";
 // eslint-disable-next-line import/no-unresolved
 import { run } from "chromatic/node";
 import { promisify } from "util";
 
-import { BUILD_STARTED, START_BUILD, UPDATE_PROJECT, UpdateProjectPayload } from "./constants";
+import {
+  BUILD_STARTED,
+  READ_PROJECT,
+  RESPOND_PROJECT,
+  START_BUILD,
+  UPDATE_PROJECT,
+  UpdateProjectPayload,
+} from "./constants";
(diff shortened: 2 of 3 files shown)


tmeasday:
Yeah, it makes sense. Let's discuss what we think the best approach is for details from an addon that:

a) any dev server wants the latest value for at startup
b) there may be changes to down the track that need to be polled and pushed to the client.

- (b) might not be necessary for project id, but it will be for branch/commit/uncommitedHash.

If you think about it (a) + (b) are the exact problem that we have with the story index generator/`index.json`[1], and there we:
 1. Have a HTTP endpoint (`/index.json`) for dev builds
 2. Publish an event (`STORY_INDEX_INVALIDATED`) on the server channel when it changes
 3. Also build the same file (`/index.json`) statically for prod builds.

The HTTP endpoint (1) is nice because it mirrors (3) without any extra work (the client doesn't need to know the difference).

Now, in this addon's case, we don't need 3, but if we are thinking about a general mechanism for all addons, I wonder if the above mechanism is the way to do it. Something like:

```ts
export const dataEndpoints = (endpoints) => {
 return {
    '/visual-tests.json': {
       setData: async (callback) => {
          let data = getProjectIdCommitEtc();
		  callback(data);

          setInterval(() => {
             let newData = getProjectIdCommitEtc();
             if (data !== newData) {
               data = newData;
               callback(data);
             }
          }, 1000);
       },
    },
  }
});
```
We could even use a generator API to be super clean about it.

[1] We use a file watcher rather than polling but that's a minor detial.

tmeasday:
@ndelangen I chatted to @shilman about this a bit and he connected this use case to the more general/longer term use case of the annotations server. If you think about it, there are various server-side calculated annotations that both SB core and various addons might create:

  - `/index.json` - Created by core, applies to stories (also the basis for the list of stories), needed for built+dev SBs, can change over time.
  - `/project.json` - Created by core, applies at the project level, not really "needed" (by manager/preview) but makes sense for dev+built SBs, can change over time (if you consider fields like `storyCount`)
  - Component+story annotations - Created by docs/annotations addon, applies at the story level, needed for built+dev SBs, can change over time.
  - Visual Tests addon info - Created by addon, applies at the project level, dev only, can change over time.
  
  
With the above in mind, maybe we should start on some kind of "annotations" API:

 - the ability for an addon to register metadata on a namespace at the project level and invalidate it (so similar to my [suggestion above](https://github.com/chromaui/addon-visual-tests/pull/15#issuecomment-1636698585) but just a sub key of `project.json`) 
 
  - a manager side API to get project-level annotation that's reactive: `api.getProjectAnnotions()` and `api.onProjectAnnotationsInvalidated()` or even a generator. WDYT?

tmeasday:
This is out of date, but we should still address this issue.

PR #21, opened 2023-07-27
Show reviewing errors in browser console and improve dev setup

tmeasday:
In order to test reviewing code, I made it much easier to test against dev/staging.


src/constants.ts
```diff
@@ -1,4 +1,8 @@
-export const { CHROMATIC_BASE_URL } = process.env;
+export const {
+  CHROMATIC_INDEX_URL,
+  CHROMATIC_BASE_URL = CHROMATIC_INDEX_URL || "https://www.chromatic.com",
```
ndelangen, line 3: Did you verify this works?

I guess it does because of:
https://github.com/storybookjs/storybook/blob/add71c33c6ec918d906026815d96668de1a6b344/code/builders/builder-manager/src/index.ts#L104

Long term I'd love to replace that with some better system that can take config options from the preset and even update them during dev.

ndelangen (approved):

PR #24, opened 2023-08-02
Realtime git info

ndelangen:


tmeasday (approved):
LGTM, are we planning on making a more built in mechanism for this, or just doing this for now?

ndelangen:
Yeah, long term, I'd love to make this type a thing build-in to some hook maybe.

For now, I think we should just implement it ourselves here.

PR #27, opened 2023-08-08
project picker improvements

weeksling:
https://linear.app/chromaui/issue/AP-3443/project-picker-improvements

This fixes the major UI bug that prevented the last project in a long list being displayed correctly. Automatically selected the first account to make project selection easier.

Second bullet will be split into a separate pr as it requires a change to the public API to support orderBy parameters in the queries for viewers.accounts and viewers.accounts.projects


tmeasday (approved):
LGTM [in the storybook](https://6480e1b0042842f149cfd74c-mawfvzpvyh.chromatic.com/?path=/story/screens-linkproject--select-project-many-projects)!

image

@weeksling did you look at scrolling the story in a play function? Not a big deal but a good idea if easy

weeksling:
@tmeasday added the play function to the story. It was easier than I thought

PR #34, opened 2023-08-16
Store last build ID by branch name

ghengeveld:
This should ensure we show the last local build for the _current_ branch rather than the last local build for _any_ branch.

### DIFF
--- a/src/Panel.tsx
+++ b/src/Panel.tsx
@@ -35,7 +35,7 @@ const initialGitInfo: GitInfo = {
 
 logger.debug("Initial Git info:", initialGitInfo);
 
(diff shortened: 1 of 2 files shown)


tmeasday (changes requested):
This looks good--but is there some way to refactor this to be more testable? I don't love creeping more logic into the `Panel` that is untested.

Perhaps the local storage could be a provider that could be mocked in a story for the `Panel`?

ghengeveld:
Switching this back to draft because it needs to be revised. Rather than storing the last build ID, we should retrieve the last build by branch/commit/uncommittedHash.

PR #37, opened 2023-08-17
Query `testsForStory` and `testsForStatus` separately to avoid pagination

ghengeveld:
Rather than fetching all data for all tests on the build, we only query for all data for the tests for the _currently selected story_, and additionally fetch limited data for up to 1000 (max page size) tests that are in progress, pending, denied, failed or broken. This way we avoid having to do any kind of pagination logic. 

In the unusual case that >1000 tests have such a status this will break down, but this is acceptable because it's used only to display status dots in the Storybook sidebar.

Depends on https://github.com/chromaui/chromatic/pull/7628

<details>
  <summary>📦 Published PR as canary version: <code>0.0.29--canary.37.4a8d155.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.29--canary.37.4a8d155.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.29--canary.37.4a8d155.0
  ```
</details>


tmeasday (commented):
Looks like this PR should be telescoping on https://github.com/chromaui/addon-visual-tests/pull/36#pullrequestreview-1582116572 -- a bit hard to review as a result.

But the code changes to `VisualTests.tsx` look good. I'm a bit surprised there aren't more changes to the story data though?

ghengeveld:
@tmeasday You're right I forgot to commit+push some changes. I updated the base to telescope off your PR.

src/screens/VisualTests/VisualTests.tsx
```diff
@@ -223,13 +247,12 @@ export const VisualTests = ({
...
+  const isOutdated = build && build.uncommittedHash !== gitInfo.uncommittedHash;
 
   const buildStatusUpdate =
     build &&
-    "tests" in build &&
-    testsToStatusUpdate(getFragment(FragmentTestFields, build.tests.nodes));
-
-  const isOutdated = build && build.uncommittedHash !== gitInfo.uncommittedHash;
+    "testsForStatus" in build &&
+    testsToStatusUpdate(getFragment(FragmentStatusTestFields, build.testsForStatus.nodes));
```
tmeasday, line 255: It looks like the stories just drop the `testsForStatus` fields and rely on this guard to not break. Are we OK with that? I think maybe it'd be better to include them (and bonus points for assertng `updateBuildStatus` gets called on render in a `play` function).

tmeasday (commented):
LGTM. One question about testing strategy we should address.

tmeasday (approved):

src/screens/VisualTests/VisualTests.tsx
ghengeveld: Done and done.

PR #46, opened 2023-08-23
Always show last (potentially local) build on branch

ghengeveld:
As agreed [here](https://linear.app/chromaui/issue/AP-3519#comment-e2ab9c5e).

This doesn't yet handle the case where the local commit is older than the build's `committedAt`. We'll handle that in later PR.

<details>
  <summary>📦 Published PR as canary version: <code>0.0.36--canary.46.cdee0a7.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @chromaui/addon-visual-tests@0.0.36--canary.46.cdee0a7.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.36--canary.46.cdee0a7.0
  ```
</details>

PR #48, opened 2023-08-23
attempt: clear the build id to remove it from query if gitInfo changes

weeksling:
This is not a fix for the issue so I didn't link it to the ticket. This attempted to remove the item from localstorage, which caused the text to change to "Waiting for build on ..." when I switched to main, but there were further issues. Opening draft to share in thread.

### DIFF
--- a/src/Panel.tsx
+++ b/src/Panel.tsx
@@ -65,6 +65,8 @@ export const Panel = ({ active }: PanelProps) => {
       },
       [GIT_INFO]: (info: GitInfo) => {
         setGitInfo(info);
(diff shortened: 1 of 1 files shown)


ghengeveld (approved):
I think the original issue is resolved with #46, but the build ID is still kept in localStorage for the time being. I think it makes sense to clear it on new GitInfo, even if we may end up not storing it at all in the future.

tmeasday:
Lets close this and address any further problems as part of https://linear.app/chromaui/issue/AP-3544/implement-build-of-out-date-behaviour

PR #49, opened 2023-08-24
Use local schema file for graphql-codegen

ghengeveld:
This avoids schema changes depending on what branch your local Chromatic project happens to be running. Now we just have to manually update the schema file whenever it changes (see [`lib/schema/public-schema.graphql`](https://github.com/chromaui/chromatic/blob/main/lib/schema/public-schema.graphql)). We may want to use a GitHub Action to do this automatically in the future.

I also tweaked our env vars to standardize on `CHROMATIC_BASE_URL` rather than `CHROMATIC_INDEX_URL` (which it isn't really), and added support for `CHROMATIC_API_URL` which I see being used when we start to host the API at a different subdomain.

<details>
  <summary>📦 Published PR as canary version: <code>0.0.38--canary.49.b436a70.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.38--canary.49.b436a70.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.38--canary.49.b436a70.0
  ```
</details>


tmeasday (changes requested):
Can you document how to update the schema?

ghengeveld:
> Can you document how to update the schema?

Not quite sure where to do this, because I prefer not to reference the Chromatic monorepo from the addon's source code. It's a matter of copying the file over.

tmeasday:
@ghengeveld I don't know why we would make it difficult for ourselves -- if we are the ones working on the addon, we should have instructions on how to do it. If the issue is that we want outside contributors to be able to work on the addon, then obscuring the place where the schema comes from doesn't really help at all.

ghengeveld:
@tmeasday I added docs on how to update the schema.

tmeasday (approved):

PR #52, opened 2023-08-25
Remove the run tests button if user is not logged in or the projectId is not set.

weeksling:
This adds a simple check to remove the run tests button from the view until the user has logged in and a projectId is set.

Had to implement a new version of the useAddonState (really useSharedState) hook from the [storybook source code](https://github.com/storybookjs/storybook/blob/4d509a51811427e625f619f359db46c8117eee65/code/lib/preview-api/src/modules/store/hooks.ts#L47). This may be able to be pushed back up to the storybook repo, but there may be further updates to do so. In particular:
 - to make this work in the addon, we needed to use the manager api, not the preview-api. For some reason the hook in storybook is inside of preview-api. I think this is a mistake, but not sure.
 - I added the client events to broadcast the changes to any other instance of the hook. This doesn't solve for making this accessible on the server side. This would require some other mechanism to store the value outside of react/hooks. This might be related to this existing set/getAddonState method in [storybook manager/addons api](https://github.com/storybookjs/storybook/blob/4d509a51811427e625f619f359db46c8117eee65/code/lib/manager-api/src/modules/addons.ts#L64C4-L64C4) but not clear how we would adapt this to work with react (perhaps just set both and mirror the state?)

<details>
  <summary>📦 Published PR as canary version: <code>0.0.42--canary.52.d677995.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.42--canary.52.d677995.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.42--canary.52.d677995.0
  ```
</details>


weeksling:
Note: This only accounts for the reactivity of the run tests button since it involved this new hook. Will be pushing further changes for the actual build status and errors. Wanted to get a review here and opinions on the storybook side of things. I also added a branch to storybook repo, but was having problems with husky and used --no-verify, so there may be some git-pre-commit hook work that was skipped. I haven't tested that draft yet, and would like to get opinions before trying to.

ndelangen:
The hook is in both.

If you're adding something to the storybook manager you need to use the manager-api.

If you use the preview-api in the manager it will cause tremendous bloat in your addon, and possibly not work at all.

The implementation of your useAddonState seems like the one for the preview-api to me.
That seems wrong.

The useSharedState is not for public use, the useAddonState is.

tmeasday:
Can we discuss here or in slack what the issues are with SB's existing `useAddonState` (I know @ghengeveld had some) and see if we can collectively figure out the best way forward? It seems like this isn't quite it just yet, according to @ndelangen.

ndelangen:
I opened a PR here:
https://github.com/chromaui/addon-visual-tests/pull/54

... changing it to the `useAddonState` from `@storybook/manager-api`.
Looks to me like it's working as designed.

What were the issues discovered?
How do I reproduce the problem?

src/components/RunTestsButton.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { RunTestsButton } from "./RunTestsButton";
+
+const meta = {
+  component: RunTestsButton,
+  args: {
+    isStarting: false,
+    projectId: "projectId",
+    accessToken: "accessToken",
+    startBuild: action("startBuild"),
+    key: TOOL_ID,
```
tmeasday, line 14: I don't think you are allowed to have a prop called `key` in React?

src/components/RunTestsButton.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { TOOL_ID } from "../constants";
+import { RunTestsButton } from "./RunTestsButton";
+
+const meta = {
+  component: RunTestsButton,
+  args: {
+    isStarting: false,
+    projectId: "projectId",
+    accessToken: "accessToken",
```
tmeasday, line 12: Can we pass in `isLoggedIn` or similar? Does the button need to know the accessToken?

src/components/RunTestsButton.stories.tsx
tmeasday: I don't believe this would be necessary any more anyway, I think the `key` prop was just there to distinguish otherwise identical `IconButtons`

tmeasday (changes requested):
Looks good, I have some questions

src/components/RunTestsButton.stories.tsx
weeksling: It could be! I used the prop directly, but renaming the prop and passing !!accessToken would be cleaner

src/components/RunTestsButton.stories.tsx
weeksling: Yes I meant to remove this

tmeasday (approved):

tmeasday:
@ndelangen we merged this in the interests of progress but let's continue on your PR as a fast follow up.

PR #64, opened 2023-08-31
First pass at server compatible addon state

tmeasday:
Fixes AP-3557, AP-3544

Prototype `useAddonState` that is server-compatible, and use for state-based manager-server communication (git info, and project info).

To QA:

**Git state**

1. Check the right branch is selected on startup
2. Change branch / commit, check it is reflected in the UI immediately
3. Refresh browser, check it displays the updated branch/commit.

**Project Info**

1. Comment out the [addon's options in main.ts](https://github.com/chromaui/addon-visual-tests/blob/c686270c7f22b9996f50faafd20094350a4b63f6/.storybook/main.ts#L42)
4. Run, pick a project, check you see instructions (as writing failed), check you can keep going
5. Comment out the options and set name: "../src/dev.ts"
6. Run, pick a project, check you see success screen, check project id/token were written, check you can keep going.
7. Start SB with options already in there, check everything works.

<details>
  <summary>📦 Published PR as canary version: <code>0.0.51--canary.64.8673acc.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.51--canary.64.8673acc.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.51--canary.64.8673acc.0
  ```
</details>

### DIFF
--- a/src/useAddonState/common.ts
+++ b/src/useAddonState/common.ts
@@ -1,0 +1,10 @@
+export const getValue = `experimental_useAddonState_setValue`;
+export type GetValuePayload = {
+  key: string;
+};
+
+export const setValue = `experimental_useAddonState_setValue`;
+export type SetValuePayload = {
+  key: string;
+  value: any;
+};
--- a/src/useAddonState/common.ts
+++ b/src/useAddonState/common.ts
@@ -1,4 +1,4 @@
-export const getValue = `experimental_useAddonState_setValue`;
+export const getValue = `experimental_useAddonState_getValue`;
 export type GetValuePayload = {
   key: string;
 };
--- a/src/useAddonState/manager.ts
+++ b/src/useAddonState/manager.ts
@@ -48,7 +48,7 @@ export function useAddonState<T>(key: string) {
   ensureListening(api);
 
   return [
-    apiGetAddonState(api, key),
+    apiGetAddonState<T>(api, key),
     useCallback((value: T) => apiSetAddonState(api, key, value), [api, key]),
-  ];
+  ] as const;
 }
--- a/src/useAddonState/manager.ts
+++ b/src/useAddonState/manager.ts
@@ -1,0 +1,54 @@
+import type { API } from "@storybook/manager-api";
+import { useStorybookApi } from "@storybook/manager-api";
+import { useCallback } from "react";
+
+import { getValue, GetValuePayload, setValue, SetValuePayload } from "./common";
+
+const requestedKeys = new Set<string>();
+let listening = false;
+function ensureListening(api: API) {
+  if (listening) return;
+
+  api.on(getValue, ({ key }: GetValuePayload) => {
+    const apiKey = `experimental_${key}`;
+    const value = api.getAddonState(apiKey);
+    if (value !== undefined) api.emit(setValue, { key, value } satisfies SetValuePayload);
+  });
+
+  api.on(setValue, ({ key, value }: SetValuePayload) => {
+    const apiKey = `experimental_${key}`;
+    api.setAddonState(apiKey, value);
+  });
+
+  listening = true;
+}
+
+function apiGetAddonState<T>(api: API, key: string) {
+  const apiKey = `experimental_${key}`;
+
+  // Ask the universe for the value of the key if we've never seen it before
+  if (!requestedKeys.has(key)) {
+    api.emit(getValue, { key } satisfies GetValuePayload);
+    requestedKeys.add(key);
+  }
+
+  return api.getAddonState(apiKey) as T | undefined;
+}
+function apiSetAddonState<T>(api: API, key: string, value: T) {
+  const apiKey = `experimental_${key}`;
+  api.setAddonState(apiKey, value);
+
+  api.emit(setValue, { key, value } satisfies SetValuePayload);
+}
+
+export function useAddonState<T>(key: string) {
+  const api = useStorybookApi();
+
+  // This should probably just always be happening
+  ensureListening(api);
+
+  return [
+    apiGetAddonState(api, key),
+    useCallback((value: T) => apiSetAddonState(api, key, value), [api, key]),
+  ];
+}
--- a/src/useAddonState/server.ts
+++ b/src/useAddonState/server.ts
@@ -1,0 +1,78 @@
+import type { Channel } from "@storybook/channels";
+
+import { getValue, GetValuePayload, setValue, SetValuePayload } from "./common";
+
+class AddonState {
+  private values: Record<string, any> = {};
+
+  private listeners: Record<string, ((value: any) => null)[]> = {};
+
+  get(key: string) {
+    return this.values[key]
+  }
+
+  set(key: string, value: any) {
+    this.values[key] = value;
+    (this.listeners[key] || []).forEach(l => l(value))
+  }
+
+  listen<T>(key: string, cb: (value: T) => null) {
+    this.listeners[key] = [...this.listeners[key] || [], cb];
+  }
+}
+
+const addonState = new AddonState();
+
+
+let listening = false;
+function ensureListening(channel: Channel) {
+  if (listening) return;
+
+  channel.on(getValue, ({ key }: GetValuePayload) => {
+    const value = addonState.get(key);
+    if (value !== undefined) channel.emit(setValue, { key, value } satisfies SetValuePayload);
+  });
+
+  channel.on(setValue, ({ key, value }: SetValuePayload) => {
+    addonState.set(key, value);
+  });
+
+  listening = true;
+}
+
+const requestedKeys = new Set<string>();
+function apiGetAddonState<T>(channel: Channel, key: string) {
+  // Ask the universe for the value of the key if we've never seen it before
+  if (!requestedKeys.has(key)) {
+    channel.emit(getValue, { key } satisfies GetValuePayload);
+    requestedKeys.add(key);
+  }
+
+  return addonState.get(key) as T | undefined;
+}
+
+function apiSetAddonState<T>(channel: Channel, key: string, value: T) {
+  addonState.set(key, value);
+  
+  channel.emit(setValue, { key, value } as SetValuePayload);
+}
+
+export function useAddonState<T>(channel: Channel, key: string) {
+  ensureListening(channel);
+
+  return {
+    get value() {
+      return apiGetAddonState(channel, key),
+    },
+
+    set value(newValue: T) {
+      apiSetAddonState(channel, key, newValue)
+    },
+
+    on(event:string, callback: (value:T) => null) {
+      if (event !== 'change') throw new Error('unsupported event');
+
+      addonState.listen(key, callback)
+    }
+  };
+}
--- a/src/useAddonState/server.ts
+++ b/src/useAddonState/server.ts
@@ -8,21 +8,20 @@ class AddonState {
   private listeners: Record<string, ((value: any) => null)[]> = {};
 
   get(key: string) {
-    return this.values[key]
+    return this.values[key];
   }
 
   set(key: string, value: any) {
     this.values[key] = value;
-    (this.listeners[key] || []).forEach(l => l(value))
+    (this.listeners[key] || []).forEach((l) => l(value));
   }
 
   listen<T>(key: string, cb: (value: T) => null) {
-    this.listeners[key] = [...this.listeners[key] || [], cb];
+    this.listeners[key] = [...(this.listeners[key] || []), cb];
   }
 }
 
 const addonState = new AddonState();
-
 
 let listening = false;
 function ensureListening(channel: Channel) {
@@ -53,7 +52,7 @@ function apiGetAddonState<T>(channel: Channel, key: string) {
 
 function apiSetAddonState<T>(channel: Channel, key: string, value: T) {
   addonState.set(key, value);
-  
+
   channel.emit(setValue, { key, value } as SetValuePayload);
 }
 
@@ -62,17 +61,17 @@ export function useAddonState<T>(channel: Channel, key: string) {
 
   return {
     get value() {
-      return apiGetAddonState(channel, key),
+      return apiGetAddonState(channel, key);
     },
 
     set value(newValue: T) {
-      apiSetAddonState(channel, key, newValue)
+      apiSetAddonState(channel, key, newValue);
     },
 
-    on(event:string, callback: (value:T) => null) {
-      if (event !== 'change') throw new Error('unsupported event');
+    on(event: string, callback: (value: T) => null) {
+      if (event !== "change") throw new Error("unsupported event");
 
-      addonState.listen(key, callback)
-    }
+      addonState.listen(key, callback);
+    },
   };
 }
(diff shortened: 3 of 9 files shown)


tmeasday:
This is a sketch at this stage and I am not sure it works. But the basics are as follows if you have feedback @ndelangen 

The API is as follows:

```ts
// manager/preview:
const [foo, setFoo] = useAddonState<string>('addon:foo') || 'bar';

useEffect(() => {
  setFoo('baz');
});

console.log(foo);

// server
const fooState = useAddonState<string>('addon:foo') || 'bar';

fooState.on('change', (foo:string) => {});

fooState.set('baz')

console.log(fooState.value)
```

Some notes:

1. I realised if we are setting the default to the same thing anyway, there is no point in actually transmitting it over the channel (apart from to error if it isn't the same), and in fact, to KISS, **we can just stop storing defaults entirely**. (Alternatively we could still take the default as a second argument just for type convience, but do nothing with it except for return it if not value is set).

2. The basic API is that any call will emit a `getAddonState` event to request all other channel listeners to send their latest value again. Apart from that, whenever you set a value it emits `setAddonState`, last call wins. It's pretty simple.

3. I think we can just setup single listeners to the above two events to maintain the store. I think we might as well just store keys that we've never even requested because why not?

4. The code is almost identical in the two places, we might consider refactoring to share code. But as it's manager/server/preview code that might be harder to do than is worth it, LMK your thoughts.

src/useAddonState/common.ts
```diff
@@ -0,0 +1,10 @@
+export const getValue = `experimental_useAddonState_setValue`;
```
ndelangen, line 1:
```suggestion
export const getValue = `experimental_useAddonState_getValue`;
```

src/useAddonState/manager.ts
```diff
@@ -0,0 +1,54 @@
...
+function apiGetAddonState<T>(api: API, key: string) {
+  const apiKey = `experimental_${key}`;
+
+  // Ask the universe for the value of the key if we've never seen it before
+  if (!requestedKeys.has(key)) {
+    api.emit(getValue, { key } satisfies GetValuePayload);
+    requestedKeys.add(key);
+  }
+
+  return api.getAddonState(apiKey) as T | undefined;
```
ndelangen, line 35: I imagine that `emit` won't immediately/synchronously set the value?
So, `api.getAddonState(apiKey)` will return undefined?

Maybe `apiGetAddonState` should return a `Promise`?

tmeasday: The api will return `undefined` until the response to the get call comes back from somewhere else. 

We could make it a promise in node, but we don't actually know if any other place has a value for the key so it might never resolve.

ndelangen:
@tmeasday here's a scenario:

We have state preserved in 3 places. (server=A, manager=B, preview=C).
Each has the same default state.
We could demand from the user that they specify the default state in the addon config in node.
if they define it there once, it will ensure:
- it starts the same everywhere, because they define it once
- it will be completely obvious this must be a serializable state
When a user runs `storybook dev`, A reads the default state from addon config, and injects this into the runtime of B & C

So far so good!?

Now imagine a state change happens at A
This will propagate to B & C

So far so good!?

Now imagine the user does NOT restart A, but refreshes B (and thus C)
A has a updated state, but B & C will read the default state injected into the runtime.

So that's a sync issue! State in A & B & C will be out of sync.
We can emit a "was there a state-update I missed" from B & C, and only respond if the state is not equal to default state?
But this still leaves with a time window where B & C are in the default state when they should not be.

So in dev-mode of storybook we'd actually want to pull the data from node at start, before rendering at all.
We discussed this before.

---

Here's a idea (again we have discussed this before):

When B & C start, they (both) fetch a JSON file, (we can use `project.json`?).
The contents of this JSON would contain a `Record<AddonID, AddonState>`.

When storybook is in dev-mode this JSON file is live, it's generated based on A's state.
When storybook is in build-mode, this JSON file was created with the default state from A.

Whilst the JSON file is being fetched, we have to block rendering of anything that could use the hook.

---

Once the JSON file is retrieved and the default state is set, and B & C are rendered..
...then B & C start listening on the channel for updates.

---

There's also a scenario where A & B are kept in place, but C is ejected, or refreshed.
It will re-fetch the JSON file before proceeding.

---

There's also a scenario where A is shut down or restarted, but B & C are kept running.
This is an existing situation, which should be solved somewhere in storybook.
Storybook should detect the disconnect and warn the user the server crashed/ended/is restarting.
If storybook would auto-refresh after A is finished restarting, this would fix the issue.
If the user restarts A and refreshes B after a while, things will just work again.

---

## Why a JSON file (vs using the serverChannel)?
Because that will work in both dev-mode and build-mode

## Where will this JSON content go?
We could place it in `index.json` or `project.json`, or any other .json file we already have.
We could also decide to add yet-another JSON file endpoint?
Obviously adding another JSON file to fetch would be sub-optimal for performance.
We have to ensure whichever existing file we put this in, there should be no HTTP caching on this file or a very short-lived one?.

## Why put the default state in the addon config (node)?
That would make it centralized
That would allow us to bail early if the data is not serializable
A possible downside is: you would not be able to derive default state based on something only present in the browser, like the `URL`, the `navigator`...

## Won't blocking rendering of stories until a JSON file is requested cause a performance regression?
Possibly! which is another reason why placing it in an existing file would be preferable.
The file will be requested by the B first, and then requested again by C.

## How do we block B rendering things that could use addonState?
This is relatively easy actually, because we have control over when we add/render UI added by addons.

## How do we block C rendering things that could use addonState?
StorystoreV7 is async already, this should not be hard to do.

tmeasday:
@ndelangen TLDR: I am generally in agreement (but have some discussion points), but I don't want to block this work on discussion of a feature we don't need for this addon. We can keep discussing on this PR, but perhaps (once ready) we can merge it before resolving everything?

Onto specifics:

> Whilst the JSON file is being fetched, we have to block rendering of anything that could use the hook.

I don't think we should block rendering, but we can just return `undefined` from the hook until we have the file. Otherwise we'll need to ensure we *always* fetch this file which could be a problem for cases like portable stories or the test runner. I'd rather folks just didn't assume the data is always there.

We could also delay rendering the manager UI / story until we've fetched the file (as you suggest). That'll make a lot of sense if we put the data in `index.json` (which we need before we render stories and the sidebar in the manager). 

> Obviously adding another JSON file to fetch would be sub-optimal for performance.

I don't know if that's necessarily obvious.

> We have to ensure whichever existing file we put this in, there should be no HTTP caching on this file or a very short-lived one?.

For dev mode I don't think it makes sense to cache at all, its so fast anyway. For build mode it never changes so a long lived cache is good.

---

**Where should we put it?**

We have three options:

- `index.json`
- `project.json`
- A new file, say `addons.json`

1. `index.json` would make sense if we want to block rendering to wait for it. However `index.json` is also used by things like the test runner and so bloating it could have perf implications. (How much this bloats it is pretty questionable though).

2. `project.json` is not currently fetched by the manager/preview and so I'd probably say not.

3. A new file could be fine but does add a bit of extra complexity.

tmeasday:
@ndelangen - this is probably mergeable as is. LMK your thoughts on my changes to use state rather than events; I think it's easier to understand and likely less buggy.

We should have more tests for `useAddonState` before merging to SB (and a preview implementation) but I think this proves the concept and is enough for the addon for now.

If you want to test the project id updating (you probably should), do the following:

1. Comment out the [addon's `options` in `main.ts`](https://github.com/chromaui/addon-visual-tests/blob/c686270c7f22b9996f50faafd20094350a4b63f6/.storybook/main.ts#L42)
2. Run, pick a project, check you see instructions (as writing failed), check you can keep going
3. Comment out the options and set `name: "../src/dev.ts"`
4. Run, pick a project, check you see success screen, check project id/token were written, check you can keep going.
5. Start SB with options already in there, check everything works.
6.

src/useAddonState/manager.ts
```diff
@@ -0,0 +1,54 @@
...
+  const apiKey = `experimental_${key}`;
+  api.setAddonState(apiKey, value);
+
+  api.emit(setValue, { key, value } satisfies SetValuePayload);
+}
+
+export function useAddonState<T>(key: string) {
+  const api = useStorybookApi();
+
+  // This should probably just always be happening
```
ghengeveld, line 47: What does this comment mean?

src/useAddonState/manager.ts
```diff
@@ -0,0 +1,54 @@
...
+    requestedKeys.add(key);
+  }
+
+  return api.getAddonState(apiKey) as T | undefined;
+}
+function apiSetAddonState<T>(api: API, key: string, value: T) {
+  const apiKey = `experimental_${key}`;
+  api.setAddonState(apiKey, value);
+
+  api.emit(setValue, { key, value } satisfies SetValuePayload);
```
ghengeveld, line 41: Why does this need both a set and an emit? To tell listeners on other sides of the channel about the update? This should be explained somewhere.

src/useAddonState/manager.ts
```diff
@@ -0,0 +1,54 @@
...
+  const api = useStorybookApi();
+
+  // This should probably just always be happening
+  ensureListening(api);
+
+  return [
+    apiGetAddonState<T>(api, key),
+    useCallback((value: T) => apiSetAddonState(api, key, value), [api, key]),
+  ] as const;
+}
```
ghengeveld, line 54: This file needs some tests.

src/useAddonState/manager.ts
```diff
@@ -0,0 +1,54 @@
...
+    api.setAddonState(apiKey, value);
+  });
+
+  listening = true;
+}
+
+function apiGetAddonState<T>(api: API, key: string) {
+  const apiKey = `experimental_${key}`;
+
+  // Ask the universe for the value of the key if we've never seen it before
```
ghengeveld, line 29: Perhaps also clarify that we're listening for an answer, which may take a while and may never come.

src/useAddonState/manager.ts
ghengeveld: I agree with @ndelangen, better to have an API that returns a Promise, than one that (synchronously) says "no" but then says "yes" when called again. A timeout can avoid the never-resolving issue.

src/useAddonState/server.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  // for testing
+  reset() {
+    this.values = {};
+    this.listeners = {};
+  }
+}
+
+const addonState = new AddonState();
```
ghengeveld, line 30: Why use a class at all? I don't see how it's beneficial.

src/useAddonState/server.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  channel.on(getValue, ({ key }: GetValuePayload) => {
+    const value = addonState.get(key);
+    if (value !== undefined) channel.emit(setValue, { key, value } satisfies SetValuePayload);
+  });
+
+  channel.on(setValue, ({ key, value }: SetValuePayload) => addonState.set(key, value));
+
+  listening = true;
+}
```
ghengeveld, line 44:
```suggestion
type ChannelLike = Pick<Channel, "emit" | "on">

let ensureListening = (channel: ChannelLike) => {
  channel.on(getValue, ({ key }: GetValuePayload) => {
    const value = addonState.get(key);
    if (value !== undefined) channel.emit(setValue, { key, value } satisfies SetValuePayload);
  });

  channel.on(setValue, ({ key, value }: SetValuePayload) => addonState.set(key, value));

  ensureListening = (channel: ChannelLike) => {};
}
```

src/useAddonState/server.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const requestedKeys = new Set<string>();
+// For testing
+export function resetAddonState() {
+  addonState.reset();
+  requestedKeys.clear();
+  listening = false;
+}
+
+function apiGetAddonState<T>(channel: Pick<Channel, "emit" | "on">, key: string) {
```
ghengeveld, line 54:
```suggestion
function apiGetAddonState<T>(channel: ChannelLike, key: string) {
```

src/useAddonState/server.ts
```diff
@@ -0,0 +1,88 @@
...
+  // Ask the universe for the value of the key if we've never seen it before
+  if (!requestedKeys.has(key)) {
+    channel.emit(getValue, { key } satisfies GetValuePayload);
+    requestedKeys.add(key);
+  }
+
+  return addonState.get(key) as T | undefined;
+}
+
+function apiSetAddonState<T>(channel: Pick<Channel, "emit" | "on">, key: string, value: T) {
```
ghengeveld, line 64:
```suggestion
function apiSetAddonState<T>(channel: ChannelLike, key: string, value: T) {
```

src/useAddonState/server.ts
```diff
@@ -0,0 +1,88 @@
...
+  addonState.set(key, value);
+
+  channel.emit(setValue, { key, value } as SetValuePayload);
+}
+
+export function useAddonState<T>(channel: Pick<Channel, "emit" | "on">, key: string) {
```
ghengeveld, line 70:
```suggestion
export function useAddonState<T>(channel: ChannelLike, key: string) {
```

src/useAddonState/server.ts
```diff
@@ -0,0 +1,88 @@
...
+  ensureListening(channel);
+
+  return {
+    get value() {
+      return apiGetAddonState(channel, key);
+    },
+
+    set value(newValue: T) {
+      apiSetAddonState(channel, key, newValue);
+    },
```
ghengeveld, line 80: Why the getter/setter? I'm not digging the `addonState.value = ...` syntax personally (it's just preference though).

ghengeveld (commented):
I assume `useAddonState` is going to make its way back to Storybook?

Overall it took me longer to grok this code than it probably should have. A good dose of explanatory comments / JSDoc would be very welcome.

src/useAddonState/manager.ts
tmeasday: But you can't do anything with a promise in a React component, can. you? And for consistency would you always have to return a promise?

We could look at using suspense or something but that seem complicated to me. This feels more idiomatic for a React hook.

src/useAddonState/manager.ts
tmeasday: Yeah, it's pretty much the same code as the server version, so I didn't bother reproducing the tests (and doing componentized versions like @ndelangen has done in https://github.com/storybookjs/storybook/pull/23991) until we agree on the details.

src/useAddonState/server.ts
tmeasday: I guess no particular reason. It lets you have private members?

src/useAddonState/server.ts
tmeasday: Very happy to change the API, I'm to wedded to it at all.

src/useAddonState/manager.ts
tmeasday: I think probably in SB we want to setup one copy of these listeners when the manager boots up and just always write values to the store/respond to requests for values, rather than only do it "on demand".

PR #65, opened 2023-08-31
Clear storyStatus when no longer relevant

ndelangen:
# What I did:

- clear storyStatus when we get partial buildData
- clear storyStatus when there is no build

<details>
  <summary>📦 Published PR as canary version: <code>0.0.53--canary.65.575bce7.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.53--canary.65.575bce7.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.53--canary.65.575bce7.0
  ```
</details>


tmeasday (commented):
LGTM. Is it possible to do some assertion a play function that we call the right version of `updateStatus` at the right time in the a `VisualTests` story?

package.json
```diff
@@ -70,19 +70,19 @@
     "@graphql-codegen/cli": "^4.0.1",
     "@graphql-codegen/client-preset": "^4.0.1",
     "@graphql-typed-document-node/core": "^3.2.0",
-    "@storybook/addon-actions": "7.4.0",
-    "@storybook/addon-essentials": "7.4.0",
-    "@storybook/addon-interactions": "7.4.0",
-    "@storybook/addon-mdx-gfm": "7.4.0",
-    "@storybook/channels": "7.4.0",
-    "@storybook/client-logger": "7.4.0",
+    "@storybook/addon-actions": "0.0.0-pr-24007-sha-c31c498a",
```
tmeasday, line 73: Do we have a sense when this API will be available? I'm not sure if we want to be stuck on a canary for a while if we are giving this to customers and the SB release process is back to 6 weeks.

Can we work around (i.e. do a bit more on our end) while the API is not yet available?

src/screens/VisualTests/VisualTests.stories.tsx
```diff
@@ -210,6 +211,19 @@ export const NoBuild: Story = {
...
     ),
   },
+  render: ({ ...args }) => {
+    return (
+      <VisualTests
+        {...args}
+        updateBuildStatus={(fn) => args.updateBuildStatus(typeof fn === "function" ? fn({}) : fn)}
+      />
+    );
+  },
```
tmeasday, line 221: Wouldn't it be simpler to just do 

```ts
const updateBuildStatusAction = action('updateBuildStatus');
args: {
   updateBuildStatus: jest.fn((fn) => updateBuildStatusAction(typeof fn...))
}
```

tmeasday (commented):
Great job on the tests ✅ I have a couple questions before we merge this though

src/screens/VisualTests/VisualTests.tsx
```diff
@@ -266,7 +273,14 @@ export const VisualTests = ({
...
   useEffect(() => {
-    if (buildStatusUpdate) updateBuildStatus(buildStatusUpdate);
+    if (buildStatusUpdate) {
+      updateBuildStatus((state) => ({
+        ...createEmptyStoryStatusUpdate(state),
+        ...buildStatusUpdate,
+      }));
+    } else {
+      updateBuildStatus(createEmptyStoryStatusUpdate);
+    }
```
ghengeveld, line 283:
```suggestion
    updateBuildStatus((state) => ({
      ...createEmptyStoryStatusUpdate(state),
      ...buildStatusUpdate,
    }));
```
This should do the same thing, right? Spreading null/undefined is valid syntax.

src/screens/VisualTests/VisualTests.stories.tsx
ghengeveld: I also don't really see why a render method is necessary. Even if it is, I'm not sure its worthwhile.

ghengeveld (approved):

src/screens/VisualTests/VisualTests.stories.tsx
```diff
@@ -1,8 +1,9 @@
 import { action } from "@storybook/addon-actions";
-import { expect } from "@storybook/jest";
+import { expect, jest } from "@storybook/jest";
```
ghengeveld, line 2: `jest` may or may not be unused?

ndelangen:
@gert the custom render method is easier way to override the arg.
You suggested adding a arg with a `_` prefix, remember? I think this is a much much cleaner solution.

In fact the `_` prefix method didn't work, this one does.

src/screens/VisualTests/VisualTests.stories.tsx
ndelangen: Would be nice if this was an error in linting or type checking..

src/screens/VisualTests/VisualTests.stories.tsx
ndelangen: OK, an approach similar enough to @tmeasday suggestion worked.

package.json
ndelangen: It's in an alpha now, upgrading...

tmeasday:
@ndelangen let me try

tmeasday:
@ndelangen this approach works although has some typing issues that may be resolvable:

```ts
  args: {
    updateBuildStatus: action("updateBuildStatus"),
  },
  play: async ({ args }) => {
    await waitFor(() => {
      expect(args.updateBuildStatus).toHaveBeenCalled();
      const fn = args.updateBuildStatus.mock.calls[0][0];
      expect(fn({})).toEqual({});
    });
  },
```

WDYT?

ndelangen:
I think the custom render approach is a simpler approach @tmeasday

PR #66, opened 2023-08-31
Notify users when a build is completed or when a build fails

weeksling:
This adds a notification on error or on completion of a build. It adds an onTaskError callback to the node-cli and uses it to push error events to the addon.

Note: The errors returned from the CLI are quite long and poorly formatted for our notifications. I've opted here to show only a generic message pointing them to the terminal for now. Copy subject to change.

Also note: This depends on an updated version of chromatic-cli with the new option for onTaskError https://github.com/chromaui/chromatic-cli/pull/808


weeksling:
Marked as draft as this depends on an unmerged cli pr and the tests are currently failing. 
Feedback on the format of the event or how else to display the error is appreciated.

For now, we will point them to check the terminal and make notes of the types of errors that we see often.

Errors are also logged to console via storybook's logger.error method. This currently logs the originalError, but could be replaced with the formattedError if that's more appropriate.

 - Also of note: the output from the cli in the storybook-server terminal gets interrupted often by output from the addon and from storybook. Will be important to test if this also interrupts our error reporting.

Will update tomorrow with screenshots of the notifications to verify against (no clear way to document these with stories).

It will create notifications:
 - when there's an error in the build
 - when the build completes

There will not be a notification when a build starts as that will be handled by the separate progress/status story as well as the button's loading state.

tmeasday (approved):
@weeksling looks like this is a telescoping PR off of mine, perhaps change the base?

The content of your commit looks good though: https://github.com/chromaui/addon-visual-tests/pull/66/commits/cf4bb30e52615776d8db467f427683c873d61d91 ✅

Regarding the test failures, I think maybe you just need to publish a canary of the CLI and reference that in `package.json`? Can walk you through how to do that if you don't know. Or we can just wait for your CLI PR to get released.

Apart from that, this looks good. Screenshots would be good, I guess this is a bit of a tricky one to get stories for.

My only question is about logging the error- doesn't our CLI already log it? So isn't that just going to lead to it being logged twice?

weeksling:
@tmeasday It is telescoping off your pr for the status update to notify on completion. I had the base as yours but switched it because I had rebased this pr with main and a lot of unexpected changes were showing up. We could just bring both up to date.

Regarding the logging that makes sense. I had thought the logger would report to browser console as well for users to see, but if it a client logger also outputs to the terminal we could remove that.

I'll update this with those changes.

I'm open to publishing a canary build, but if the changes I made to the other pr are acceptable, I think we could merge soon.

PR #69, opened 2023-09-01
Separate running/latest build from viewed stories

tmeasday:
Implement behaviour for "next" builds discussed in linear, namely:

1. Don't switch the story we are looking until you click a button in the eyebrow
2. But if you change story, switch to a "next" latest build as soon as it is available.
3. Additionally, always use the next build to drive the sidebar.

Relies on https://github.com/chromaui/chromatic-cli/pull/810 and telescoping on #68

<details>
  <summary>📦 Published PR as canary version: <code>0.0.51--canary.69.503ff0c.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.51--canary.69.503ff0c.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.51--canary.69.503ff0c.0
  ```
</details>


src/screens/VisualTests/VisualTests.tsx
```diff
@@ -261,26 +288,48 @@ export const VisualTests = ({
...
 
-  if (!build || error) {
+    // We always want storyBuildInfo set if we can
+    if (!storyBuildInfo) {
+      setStoryBuildInfo({ storyId, buildId: nextBuild.id });
+
+      // If you change story, we reset storyBuildInfo to the next build (you see the next build)
+    } else if (storyBuildInfo.storyId !== storyId) {
+      setStoryBuildInfo({ storyId, buildId: nextBuild.id });
+    }
```
ghengeveld, line 323: This seems inefficient, `setStoryBuildInfo` updates `storyBuildInfo` which is a dependency of the useEffect. I think we need to use the setState callback variant here. It also seems both cases set the same value?

```js
setStoryBuildInfo((storyBuildInfo) => {
  return (!storyBuildInfo || storyBuildInfo.storyId !== storyId)
    ? { storyId, buildId: nextBuild.id }
    : storyBuildInfo;
})
```

This way you can drop `storyBuildInfo` from the deps array.

src/screens/VisualTests/VisualTests.tsx
```diff
@@ -261,26 +288,48 @@ export const VisualTests = ({
...
+  }, [storyBuildInfo, hasNextBuild, nextBuild?.id, storyId]);
+
+  const switchToNextBuild = useCallback(
+    () => setStoryBuildInfo({ storyId, buildId: nextBuild.id }),
```
ghengeveld, line 327: `nextBuild` can be null/undefined.

ghengeveld (approved):
I think we're going to refactor this a bit more later on (mostly to pull apart `VisualTests.tsx`), but this is a step in the right direction 👍

src/Panel.tsx
```diff
@@ -34,28 +34,9 @@ export const Panel = ({ active, api }: PanelProps) => {
...
-          localStorage.setItem(DEV_BUILD_ID_KEY, id);
-        }
-        if (step === "snapshot" || step === "complete") {
-          setIsStarting(false);
-        }
-      },
-    },
-    []
-  );
+  const [runningBuild] = useAddonState<RunningBuildPayload>(RUNNING_BUILD);
```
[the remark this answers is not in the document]
tmeasday: It'll be `undefined` until we get the real current state back from the manager thanks to `useAddonState`. It's very fast, so really we just need to make sure we don't error while it's undefined.

src/screens/VisualTests/VisualTests.tsx
```diff
@@ -376,12 +383,11 @@ export const VisualTests = ({
...
     (runningBuild && runningBuild.step !== "complete") ||
     // Even if there's no build running, we want to show the next build if it hasn't been selected.
-    (nextBuild && nextBuild.id !== storyBuild?.id);
+    (canSwitchToNextBuild && nextBuild.id !== storyBuild?.id);
   const buildStatus = showBuildStatus && (
     <BuildProgress
       runningBuild={runningBuild}
-      nextBuild={nextBuild}
-      switchToNextBuild={switchToNextBuild}
+      switchToNextBuild={canSwitchToNextBuild && switchToNextBuild}
```
ghengeveld, line 392: I don't think `canSwitchToNextBuild` is needed here, since `switchToNextBuild` also checks it.

tmeasday: The `BuildProgress` component uses it to tell if it should display the link or not.

See:
 - https://www.chromatic.com/test?appId=6480e1b0042842f149cfd74c&id=64f567c8beb8a0bf7bd418de vs
 - https://www.chromatic.com/test?appId=6480e1b0042842f149cfd74c&id=64f567c8beb8a0bf7bd418df

PR #70, opened 2023-09-04
Fix status title and viewport/browser selector for accepted stories

ghengeveld:
Changes the title to "No changes" and hides the status dots on the viewport and browser selector when changes are accepted.

Also fixes AP-3601

<details>
  <summary>📦 Published PR as canary version: <code>0.0.52--canary.70.444e17c.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.52--canary.70.444e17c.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.52--canary.70.444e17c.0
  ```
</details>

### DIFF
--- a/src/components/ViewportSelector.tsx
+++ b/src/components/ViewportSelector.tsx
@@ -1,7 +1,7 @@
 import { Icon } from "@storybook/design-system";
 import React from "react";
 
-import { ComparisonResult, ViewportInfo } from "../gql/graphql";
+import { ComparisonResult, TestStatus, ViewportInfo } from "../gql/graphql";
 import { aggregateResult } from "../utils/aggregateResult";
 import { ArrowIcon } from "./icons/ArrowIcon";
 import { StatusDot, StatusDotWrapper } from "./StatusDot";
@@ -10,12 +10,14 @@ import { TooltipMenu } from "./TooltipMenu";
 type ViewportData = Pick<ViewportInfo, "id" | "name">;
 
 interface ViewportSelectorProps {
+  testStatus: TestStatus;
   selectedViewport: ViewportData;
   onSelectViewport: (viewport: ViewportData) => void;
   viewportResults: { viewport: ViewportData; result: ComparisonResult }[];
 }
 
 export const ViewportSelector = ({
+  testStatus,
   selectedViewport,
   viewportResults,
   onSelectViewport,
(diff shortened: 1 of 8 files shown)


src/screens/VisualTests/StoryInfo.tsx
```diff
@@ -92,7 +92,9 @@ export const StoryInfo = ({
     details = (
       <Text>
         <b>
-          {changeCount ? pluralize("change", changeCount, true) : "No changes"}
+          {changeCount && status === TestStatus.Pending
+            ? pluralize("change", changeCount, true)
+            : "No changes"}
```
tmeasday, line 97: I know this was what was requested but it doesn't quite make sense to me. The change doesn't go away when it is accepted?

Should it be still there for denied btw?

ghengeveld: Yeah I felt the same way. "Accepted 1 change" would be more sensible, I think. @MichaelArestad WDYT?

src/components/ViewportSelector.tsx
```diff
@@ -10,12 +10,14 @@ import { TooltipMenu } from "./TooltipMenu";
 type ViewportData = Pick<ViewportInfo, "id" | "name">;
 
 interface ViewportSelectorProps {
+  testStatus: TestStatus;
```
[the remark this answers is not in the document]
ghengeveld: I'll change it to `isAccepted`, to avoid any confusion.

src/utils/summarizeTests.ts
```diff
@@ -2,12 +2,12 @@ import { ComparisonResult, StoryTestFieldsFragment, TestResult, TestStatus } fro
 import { aggregateResult } from "./aggregateResult";
 
 function pickStatus(statusCounts: { [K in TestStatus]?: number }) {
+  if (statusCounts[TestStatus.InProgress] > 0) return TestStatus.InProgress;
   if (statusCounts[TestStatus.Failed] > 0) return TestStatus.Failed;
```
tmeasday, line 6: Maybe Failed should be before InProgress at least? That's the way it works on the Chromatic side: https://github.com/chromaui/chromatic/blob/6da7d96b3c7eaa8383616bee2665cd54a118d6fc/services/index/model/Test.ts#L425-L433

tmeasday (approved):
LGTM

src/screens/VisualTests/StoryInfo.tsx
MichaelArestad: Those are good points. The idea was to say there are no changes left for a user to review. Do you have a better idea as to what we could show here?

ghengeveld: "Accepted 1 change" or just "Changes accepted" are both good options I think.

MichaelArestad: Will it show "Changes accepted" the next time they arrive at this view? Will other users see this?

tmeasday: I think the simplest thing would be for everyone to see the same thing.

PR #56, opened 2023-08-29
Build progress UI

tmeasday:
Telescoping on https://github.com/chromaui/addon-visual-tests/pull/55

A very initial implementation of the progress eyebrow.

PR #76, opened 2023-09-06
[AP 3560] Build UI for build status

weeksling:
This builds out most of the ui for the progress bar and build status, extending https://github.com/chromaui/addon-visual-tests/pull/56

Note this is not complete.
 - The onTaskCompleted callback does not work as intended, and it sets step to undefined until build.
 - The entire index side of this will change with recent changes so it needs to use whatever the latest is for passing in progress.
 - Steps in the cli do not accurately match the tasks in the cli. There are several steps in the design for the ui based on the actual cli output. These steps need to be added to the types possible on the   BuildProgressPayload and in the actual cli.
 - This doesn't include the concept of stepsHistory yet. We need to add a useEffect that will store/update this value onCompletion of a step. This could also be passed as a prop.

<details>
  <summary>📦 Published PR as canary version: <code>0.0.68--canary.76.d059219.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.68--canary.76.d059219.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.68--canary.76.d059219.0
  ```
</details>


src/components/BuildProgressLabel.tsx
```diff
@@ -1,21 +1,17 @@
...
-  <>{messageMap[runningBuild.step](runningBuild)}</>
-);
+export const BuildProgressLabel = ({ runningBuild }: BuildProgressLabelProps) => {
+  const { emoji, renderProgress } = BUILD_STEP_CONFIG[runningBuild.currentStep];
+  return (
+    <>
+      {emoji} {renderProgress(runningBuild)}
+    </>
+  );
+};
```
tmeasday, line 17: Would stories for these be helpful?

src/index.ts
```diff
@@ -48,6 +51,120 @@ const observeGitInfo = async (
   return () => clearTimeout(timer);
 };
 
+const getConfig = (
```
tmeasday, line 54: This seems like a misnomer

src/index.ts
```diff
@@ -1,21 +1,24 @@
+/* eslint-disable no-param-reassign */
```
tmeasday, line 1: All the code in this file seems long and complex enough to warrant a separate file and a set of tests.

src/index.ts
```diff
@@ -48,6 +51,120 @@ const observeGitInfo = async (
...
+      if (progress && total) {
+        newPercentage += stepPercentage * (progress / total);
+      }
+
+      if (!["upload", "snapshot"].includes(ctx.task)) {
+        const { estimateDuration } = BUILD_STEP_CONFIG[ctx.task];
+        const stepIndex = BUILD_STEP_ORDER.indexOf(ctx.task);
+        newPercentage =
+          Math.max(newPercentage, buildProgressPercentage) +
+          (2000 / estimateDuration) * stepPercentage;
```
tmeasday, line 103: I assume this is the same constant as the 2000 below? can we pull it into a constant please.

src/index.ts
```diff
@@ -48,6 +51,120 @@ const observeGitInfo = async (
...
+
+        setTimeout(() => {
+          // Intentionally reference the _current_ value here (after timeout)
+          const { currentStep } = runningBuildState.value;
+          const index = BUILD_STEP_ORDER.indexOf(currentStep as any);
```
tmeasday, line 108: Why the `as any`? Can't you use the `isKnownStep` check and short circuit if not?

src/index.ts
```diff
@@ -48,6 +51,120 @@ const observeGitInfo = async (
...
+        ctx.task,
+        previousBuildProgress
+      );
+
+      let newPercentage = startPercentage;
+      if (progress && total) {
+        newPercentage += stepPercentage * (progress / total);
+      }
+
+      if (!["upload", "snapshot"].includes(ctx.task)) {
```
tmeasday, line 98:
```suggestion
      // If the step doesn't have a progress event, simulate one by synthetically updating progress
      // every CONSTANT_NAME ms
      if (!["upload", "snapshot"].includes(ctx.task)) {
```

tmeasday: Also, perhaps we should add `hasProgressEvent` to `buildSteps.ts` rather than hardcoding it here?

src/screens/VisualTests/BuildEyebrow.stories.ts
```diff
@@ -80,26 +93,12 @@ export const Snapshot: Story = {
...
-};
-
-export const LatestNeedToUpdate: Story = {
-  args: {
-    switchToNextBuild: undefined,
-  },
-  parameters: withFigmaDesign(
-    "https://www.figma.com/file/GFEbCgCVDtbZhngULbw2gP/Visual-testing-in-Storybook?type=design&node-id=2127-448761&mode=design&t=70EtYCn1H7hB8PAk-0"
-  ),
-};
```
tmeasday, line 105: We still need these two stories I think -- you updated the code for them after all:

image

src/screens/VisualTests/BuildProgress.tsx
```diff
@@ -20,46 +38,90 @@ export const Bar = styled.div<{ percentage: number }>(({ theme, percentage }) =>
...
-  const text = runningBuild ? (
-    <BuildProgressLabel runningBuild={runningBuild} />
-  ) : switchToNextBuild ? (
-    "There's a newer snapshot with changes"
-  ) : (
-    "Reviewing is disabled because there's a newer snapshot on <branch>"
-  );
+  useEffect(() => {
+    stepHistory.current[buildProgress.currentStep] = { ...buildProgress };
+  }, [buildProgress]);
```
tmeasday, line 95: Does this approach work if you change story or close+reopen the addon panel mid-build?

src/screens/VisualTests/BuildProgress.tsx
```diff
@@ -20,46 +38,90 @@ export const Bar = styled.div<{ percentage: number }>(({ theme, percentage }) =>
...
+
+  const currentIndex = BUILD_STEP_ORDER.findIndex((key) => key === buildProgress.currentStep);
+  const steps = BUILD_STEP_ORDER.map((step, index) => {
+    const config = {
+      ...BUILD_STEP_CONFIG[step],
+      isCurrent: index === currentIndex,
+      isPending: index > currentIndex,
```
tmeasday, line 102: Pull these into vars rather than rewriting the guard right below?

tmeasday (changes requested):
Looks great. A lot of smaller feedback. Probably my biggest concern is the "history" being storied in a leaf component's state, that seems fragile.

src/screens/VisualTests/BuildProgress.tsx
ghengeveld: Yes, we don't remount the component when those things happen. I checked.

src/runChromaticBuild.test.ts
```diff
@@ -0,0 +1,79 @@
+// eslint-disable-next-line import/no-unresolved
+import * as chromatic from "chromatic/node";
+
+import { runChromaticBuild } from "./runChromaticBuild";
+
+jest.mock("chromatic/node", () => ({ run: jest.fn() }));
```
tmeasday, line 6:
```suggestion
import { run } from "chromatic/node";

jest.mock("chromatic/node");
const mockedRun = mocked(run);
```

tmeasday (approved):
LGTM, test seems like the right approach!

src/runChromaticBuild.test.ts
ghengeveld: If only that would work. I tried (and just tried again).

PR #86, opened 2023-09-07
Reviewing: Rerun query when accepting & un-accepting changes

ndelangen:
# What I did

As per @tmeasday's suggestion: I added a call to `rerun()` to the `onAccept` & `onUnaccept` callbacks making mutation

# How to test

- open storybook in repo
- make a change thaty will show up in the UI
- start a build & wait for it to report changes
- accept the change, assert this toggles to the reviewed state fast
- un-accept the change, assert this toggles to unreviewed state fast

<details>
  <summary>📦 Published PR as canary version: <code>0.0.62--canary.86.a78ce21.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.62--canary.86.a78ce21.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.62--canary.86.a78ce21.0
  ```
</details>

### DIFF
--- a/src/screens/VisualTests/VisualTests.tsx
+++ b/src/screens/VisualTests/VisualTests.tsx
@@ -88,7 +88,10 @@ export const VisualTests = ({
           input: { testId, status: ReviewTestInputStatus.Accepted, batch },
         });
 
-        if (reviewError) throw reviewError;
+        if (reviewError) {
+          throw reviewError;
+        }
+        rerun();
       } catch (err) {
         // https://linear.app/chromaui/issue/AP-3279/error-handling
         // eslint-disable-next-line no-console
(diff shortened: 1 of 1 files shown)


tmeasday (approved):
LGTM. Maybe we should rename the var from `rerun`? 

I tried it out and it sure did work. It does briefly show "Accept" again (I guess while the query is in flight). I guess to fix it we'd need to implement our own state var for `isReviewing` rather than relying on URQL's mutation state. WDYT?

ndelangen:
Hmmm.. I'm not sure.. it seems like there should also really be some sort of loading indicator?
@ghengeveld How do you feel we should solve this?

tmeasday:
I'm happy to move on for now and reconsider these UX nuances later for sure, btw.

ndelangen:
@tmeasday when I added `rerun()` I wondered if I should `await` it.. and I decided it should not, because `rerun` (according to types) is not an async function.

PR #90, opened 2023-09-08
Allow setting builds as args rather than parameters

tmeasday:
Trying this out - is this what you had in mind @yannbf? It seems like the types don't like it though @kasperpeulen (see the `@ts-expect-error`, not quite sure why. I can set `args.AddonVisualTestsBuild` in the meta but not in the story.

It does work though.

image

<details>
  <summary>📦 Published PR as canary version: <code>0.0.64--canary.90.6a458fa.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.64--canary.90.6a458fa.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.64--canary.90.6a458fa.0
  ```
</details>

### DIFF
--- a/src/screens/VisualTests/VisualTests.stories.tsx
+++ b/src/screens/VisualTests/VisualTests.stories.tsx
@@ -13,7 +13,14 @@ import type {
   StoryBuildFieldsFragment,
   StoryTestFieldsFragment,
 } from "../../gql/graphql";
-import { Browser, BuildStatus, ComparisonResult, TestResult, TestStatus } from "../../gql/graphql";
+import {
+  AddonVisualTestsBuildDocument,
+  Browser,
+  BuildStatus,
+  ComparisonResult,
+  TestResult,
+  TestStatus,
+} from "../../gql/graphql";
 import { AnnouncedBuild, PublishedBuild, StoryBuildWithTests } from "../../types";
 import { storyWrapper } from "../../utils/graphQLClient";
 import { playAll } from "../../utils/playAll";
@@ -127,20 +134,27 @@ const withGraphQLMutation = (...args: Parameters<typeof graphql.mutation>) => ({
   },
 });
 
-const withBuilds = ({
+const constructResult = ({
   nextBuild,
   storyBuild,
 }: {
   storyBuild?: StoryBuildFieldsFragment;
   nextBuild?: NextBuildFieldsFragment;
 }) => {
-  return withGraphQLQueryResult(QueryBuild, {
+  return {
     project: {
       name: "acme",
       nextBuild: nextBuild || storyBuild,
     },
     storyBuild,
-  });
+  };
+};
+
+const withBuilds = (args: {
+  storyBuild?: StoryBuildFieldsFragment;
+  nextBuild?: NextBuildFieldsFragment;
+}) => {
+  return withGraphQLQueryResult(AddonVisualTestsBuildDocument, constructResult(args));
 };
 
 const meta = {
@@ -148,6 +162,9 @@ const meta = {
   component: VisualTests,
   decorators: [storyWrapper],
   parameters: withBuilds({ storyBuild: passedBuild }),
+  argTypes: {
+    AddonVisualTestsBuild: { target: "msw" },
+  },
   args: {
     gitInfo: {
       userEmailHash: "abc123",
@@ -162,7 +179,11 @@ const meta = {
     setAccessToken: action("setAccessToken"),
     updateBuildStatus: action("updateBuildStatus") as any,
   },
-} satisfies Meta<typeof VisualTests>;
+} satisfies Meta<
+  Parameters<typeof VisualTests>[0] & {
+    AddonVisualTestsBuild: ResultOf<typeof QueryBuild>;
+  }
+>;
 
 export default meta;
 type Story = StoryObj<typeof meta>;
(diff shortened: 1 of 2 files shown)


src/screens/VisualTests/VisualTests.stories.tsx
```diff
@@ -341,8 +362,11 @@ export const RunningBuildInProgress: Story = {
 };
 
 export const Pending: Story = {
+  args: {
+    // @ts-expect-error
+    AddonVisualTestsBuild: constructResult({ storyBuild: pendingBuild }),
```
tmeasday, line 367: For some reason it errors if I set this arg on the `Story` but not on the `Meta`.

shilman: Does it make more sense to make more sense for the arg to be the input to `constructResult` if we want the user to be able to edit this in the controls panel?

shilman (commented):
Cool!!

src/screens/VisualTests/VisualTests.stories.tsx
tmeasday: How would that work @shilman? `constructResult` is specific to this query...

kasperpeulen: Ah, yes, if you use `typeof meta`, it only looks at the component type, and it can really do anything with the satisfies part.

You will have to do something as:

```ts
type StoryArgs = Parameters<typeof VisualTests>[0] & {
    AddonVisualTestsBuild: ResultOf<typeof QueryBuild>;
}

const meta = { } satisfies Meta<StoryArgs>;

type Story = StoryObj<StoryArgs>
```

kasperpeulen:
Really nice to see that args can be used for together with `msw`, would make it much more powerful!

src/screens/VisualTests/VisualTests.stories.tsx
kasperpeulen: That is right, you have to set them optional yourself in StoryObj.

I have thought about this scenario, but I think there is no way, as the `satisfies` constraint has no effect on the inferred return type of `meta`. 

Factories would solve this problem 🤓

tmeasday: I figured you'd say that :)

tmeasday: This worked OK, FTR: 
image

Only problem was a couple of stories that had a `render()` function, the `args` were typed wrongly :/

tmeasday:
Closing this in favour of a similar change in https://github.com/chromaui/addon-visual-tests/pull/124

Note in that PR I nested the args under `$graphql`:

https://github.com/chromaui/addon-visual-tests/blob/fcaf963a785e8a10c969d7b520054993882a825a/src/screens/VisualTests/VisualTests.stories.tsx#L281-L298

I also add a "mapping" concept:

https://github.com/chromaui/addon-visual-tests/blob/fcaf963a785e8a10c969d7b520054993882a825a/src/screens/VisualTests/VisualTests.stories.tsx#L136-L141

https://github.com/chromaui/addon-visual-tests/blob/fcaf963a785e8a10c969d7b520054993882a825a/src/screens/VisualTests/VisualTests.stories.tsx#L101-L125

It works pretty well. @shilman I think nesting under `$graphql` makes it a little more intuitive to read, but a bit more awkward to write--ie:

```ts
export const StoryAddedNotInBuild = {
  args: {
    $graphql: {
      AddonVisualTestsBuild: {
        selectedBuild: withTests({ ...pendingBuild }, []),
      },
    },
  },
}

// vs 

export const StoryAddedNotInBuild = {
  args: {
    AddonVisualTestsBuild: {
      selectedBuild: withTests({ ...pendingBuild }, []),
    },
  },
}
```

On the one hand the extra level of nesting is just more work all the time, on the other hand it makes it much more clear what this arg that has nothing to do with the component's props is doing.

shilman:
Haven't thought this through, but I wonder if we could just hoist `$graphql` and make `$x` the convention for targeted args, reducing one level of hierarchy?

tmeasday:
@shilman we could but I guess I was saying the `$graphql` is actually useful because at a glance it distinguishes the target of the args when you consider the story.

If it was `$AddonVisualTestsBuild` it would still be clear it's not a prop, but not where it's going. 

Maybe that's a good tradeoff though? I guess in any given project you are probably only going to have one or two arg "targets".

If we did it that way, we could even just make the `$-` thing a convention, given you'll have to define somewhere that `$AddonVisualTestsBuild` targets `graphql`.

shilman:
Sorry I meant the other way. Pull `$graphql` to be a peer of args.

tmeasday:
So a new annotation?

yannbf:
I think it makes sense to have them under args:

1. People already connect args to the controls table, or addons that provide a controls-like experience
2. You still maintain a scope, rather than having a freeform property in the story top annotation level
3. The $ prefix (maybe it should be $$) helps distinguish what is not a prop, the only framework that prepends certain properties with $ is angular, for observables

shilman:
Yeah a new class of annotation. Where we could have `$graphql`, `$mock`, `$<yourTargetHere>`.

I don't love it, but I'm bringing it up because we always try to be careful about scoping things correctly and then we generate these deeply nested structures that are not that ergonomic, and then we end up hoisting them later. But maybe we should consider just hoisting from the beginning under the assumption that targeted args will be a big deal and people will be using them a lot.

PR #92, opened 2023-09-11
fix: hide tooltip menu on click

weeksling:
Adds the onHide to the onClick for menu items.

@ghengeveld from our pairing session this morning.


ghengeveld (approved):
LGTM and works well.

PR #93, opened 2023-09-11
hide the toggleBaseline button when there are no changes

weeksling:
<details>
  <summary>📦 Published PR as canary version: <code>0.0.93--canary.93.565c844.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.93--canary.93.565c844.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.93--canary.93.565c844.0
  ```
</details>


src/screens/VisualTests/BuildResults.tsx
```diff
@@ -75,6 +77,9 @@ export const BuildResults = ({
     ),
   ];
 
+  const { changeCount, status } = summarizeTests(storyTests ?? []);
+  const completedNoChanges = !changeCount && status !== TestStatus.Pending;
```
tmeasday, line 81:
```suggestion
  const { changeCount, status } = summarizeTests(storyTests ?? []);
  const completedNoChanges = !changeCount && status !== TestStatus.InProgress;
```

src/screens/VisualTests/BuildResults.tsx
```diff
@@ -154,20 +159,22 @@ export const BuildResults = ({
...
+                trigger="hover"
+                hasChrome={false}
               >
-                <Icons icon="transfer" />
-              </IconButton>
-            </WithTooltip>
-          </Col>
+                <IconButton
+                  data-testid="button-toggle-snapshot"
+                  onClick={() => toggleBaselineImage()}
```
tmeasday, line 171: Do we need some logic in case you

a) select to show the baseline image
b) you switch to a story with no changes

Or does the state variable reset each time you change story?

tmeasday (changes requested):
Nice. Just chasing up a couple things

ghengeveld (changes requested):
I'm a bit confused about this change. It's hiding the button when _none_ of the tests for the story have changes. That means it will show the button if _some_ of the tests have changes, but that seems wrong to me. What if there are no changes on the currently selected test (say the one for Chrome), but there are changes to another test for the same story (say in Firefox). I thing the toggle button should reflect the currently selected **test**, not the aggregate of the currently selected **story**.

src/screens/VisualTests/BuildResults.tsx
weeksling: @tmeasday thank you for checking this. I looked at it, and the baselineImageVisible value does not reset when changing stories because the SnapshotComparison component is never unmounted. I've added some logic to reset the state of all the state variables if the storyId changes (added it as a prop as well). It works, but there is a small flicker when it resets but the new story hasn't loaded yet.

weeksling:
Note: I have updated this pr pretty significantly, but because it is closely related to the New Story UI I merged it into https://github.com/chromaui/addon-visual-tests/pull/101 to make this faster to review and merge.
Closing this for now.

PR #95, opened 2023-09-12
Add a UI state for an error on the first build

tmeasday:
@MichaelArestad I wasn't sure if there is a component I could use here. 

Also I am not sure exactly where this should link to?

<details>
  <summary>📦 Published PR as canary version: <code>0.0.99--canary.95.19f15a6.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.99--canary.95.19f15a6.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.99--canary.95.19f15a6.0
  ```
</details>


tmeasday:
@jonniebigodes any suggestion on the best place the link should go? This is when the Chromatic build errors somewhere, probably most likely the SB build itself. @MichaelArestad suggested https://www.chromatic.com/docs/setup/#troubleshooting

jonniebigodes:
@tmeasday I'm ok with it for now, as the errors are similar for each case. However, as feedback continues to roll in, I'm more than ok with updating the documentation and setting up a section, possibly named "Debugging Tests",  with some additional information and instructions on debugging it. I'm more than glad to update the addon to reflect that change. Sounds good?

weeksling (approved):

src/screens/VisualTests/NoBuild.tsx
```diff
@@ -22,59 +35,87 @@ interface NoBuildProps {
...
+  let contents;
+  if (localBuildProgress) {
+    if (localBuildProgress.currentStep === "error") {
+      const firstError = Array.isArray(localBuildProgress.originalError)
+        ? localBuildProgress.originalError[0]
+        : localBuildProgress.originalError;
+      contents = (
+        <>
+          <ErrorContainer>
+            <b>Build failed:</b> <code>{firstError?.message || "Unknown Error"}</code>{" "}
```
weeksling, line 62: Just to be sure, we want the originalError instead of the formattedError so that users see more information?

weeksling (approved):
This looks good to me. @tmeasday I had trouble reproducing this. For the case in the ticket (running a build in a repo with only one commit) now creates a new build with auto-accepted stories. Is there another cli build error you've used to test it?

tmeasday:
@weeksling hmm, yes we tweaked the CLI to allow single commit repos from the addon. You could try a repo with no commits at all (not no git setup at all, that will trigger a different issue 😉 ).

It's a while ago now, I think I figured out some `main.js` that would error for sb build but not for sb dev. Another idea that occurs to me is to just put some junk in your `build-storybook` script in `package.json`.

weeksling:
Got it, thanks! @tmeasday this formatting looks a bit off, but it looks like both the originalError and the formattedError contain the same content. I think we should \ remove the unneeded characters in the error message for the `experimental_onTaskError` formatter.
Screen Shot 2023-10-04 at 2 53 16 PM

PR #96, opened 2023-09-12
UI: Add icons in BrowserSelector list

ndelangen:


src/components/BrowserSelector.tsx
```diff
@@ -57,6 +57,7 @@ export const BrowserSelector = ({
       id: browser.id,
       onClick: () => onSelectBrowser(browser),
       right: !isAccepted && result !== ComparisonResult.Equal && <StatusDot status={result} />,
+      left: browserIcons[browser.key],
```
ghengeveld, line 60: I think `left` is deprecated in favor of `icon` (might also be the other way around).

ghengeveld (approved):

PR #99, opened 2023-09-13
Onboarding: Show auth screen when projectId is detected

ndelangen:
# What I did:

When the user opens the storybook with the projectId already setup, they don't need to opt-in to the addon anymore, thus the welcome screen should be skipped.

Unfortunately the `isSetup` flag I added is only detected later on, so I had to add a little useEffect to set the state correctly as the flag flips after the first render.

<details>
  <summary>📦 Published PR as canary version: <code>0.0.87--canary.99.f31a776.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.87--canary.99.f31a776.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.87--canary.99.f31a776.0
  ```
</details>


src/screens/Authentication/Authentication.stories.tsx
```diff
@@ -52,6 +53,15 @@ export const Welcome = {
   ),
 };
 
+export const PreSetup = {
+  parameters: withFigmaDesign(
+    "https://www.figma.com/file/GFEbCgCVDtbZhngULbw2gP/Visual-testing-in-Storybook?type=design&node-id=304-317931&t=3EAIRe8423CpOQWY-4"
+  ),
+  args: {
+    isSetup: true,
```
ghengeveld, line 61: Shouldn't this be `false` for a `PreSetup` story? Or rename the story to `PostSetup` or something.

ghengeveld: I guess maybe I'm confused as to the meaning of `isSetup`. Does it mean "the project has been setup" or "this is the setup screen" (i.e. the project _hasn't_ been setup)? It seems to be the former. Maybe change the prop name to avoid that ambiguity.

src/screens/Authentication/Authentication.tsx
```diff
@@ -8,15 +8,21 @@ import { Welcome } from "./Welcome";
...
+export const Authentication = ({ setAccessToken, isSetup }: AuthenticationProps) => {
+  const [screen, setScreen] = useState<AuthenticationScreen>(isSetup ? "signin" : "welcome");
 
   const [isMounted, setMounted] = useState(true);
   useEffect(() => () => setMounted(false), []);
+  useEffect(() => {
+    if (isSetup && screen === "welcome") {
+      setScreen("signin");
+    }
+  }, [isSetup, screen]);
```
ghengeveld, line 25: Why not something like:
```js
switch (isSetup && screen === "welcome" ? "signin" : screen) {
```
And drop the useEffect and initial state.

ghengeveld (changes requested):

src/screens/Authentication/Authentication.stories.tsx
ndelangen: "PreSetup" is short for "Previously Setup", I'll change it.
"isSetup" means there is a a projectId present, meaning someone has done that setup work already previously.

I'll change both to "hasProjectId", do you feel that avoids ambiguity?

src/screens/Authentication/Authentication.tsx
```diff
@@ -8,50 +8,45 @@ import { Welcome } from "./Welcome";
...
 
   const { onSignIn, userCode, verificationUrl } = useSignIn({
     isMounted,
     onSuccess: setAccessToken,
     onFailure: () => {},
   });
 
-  switch (screen) {
-    case "welcome":
+  switch (true) {
```
ghengeveld, line 28: Let's just make these a bunch of `if` statements then. The `switch` serves no purpose now (no fall-through, no real default).

src/screens/Authentication/SignIn.tsx
```diff
@@ -18,10 +18,12 @@ interface SignInProps {
...
-    <BackButton onClick={onBack}>
-      <BackIcon />
-      Back
-    </BackButton>
+    {onBack ? (
+      <BackButton onClick={onBack}>
+        <BackIcon />
+        Back
+      </BackButton>
+    ) : null}
```
ghengeveld, line 26:
```suggestion
    {onBack && (
      <BackButton onClick={onBack}>
        <BackIcon />
        Back
      </BackButton>
    )}
```

ghengeveld (approved):

PR #91, opened 2023-09-08
Open dialog to allow creating a new project

tmeasday:
Fixes AP-3630

Add a new item to the `LinkProject` list for "create new project" and open a dialog to the apps page of the account, which allows you to create a new project.

image

### TODO
 - [x] Get the URL from the API - https://github.com/chromaui/chromatic/pull/7714
 - [x] Link directly to the new project overlay - https://github.com/chromaui/chromatic/pull/7713
 - [x] Send a message when project is created - https://github.com/chromaui/chromatic/pull/7734
 - [x] Ensure the addon page polls underneath (Fixes AP-3620)
 - [x] Autoclose the dialog after the project is created.
 - [x] Select the project after autoclose
 - [x] Integrate the above

### TO QA

See https://github.com/chromaui/chromatic/pull/7734

<details>
  <summary>📦 Published PR as canary version: <code>0.0.83--canary.91.c5d6621.0</code></summary>
  <br />

  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @chromaui/addon-visual-tests@0.0.83--canary.91.c5d6621.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.83--canary.91.c5d6621.0
  ```
</details>

PR #100, opened 2023-09-14
Send user back to oauth flow after login

tmeasday:
Telescoping on https://github.com/chromaui/addon-visual-tests/pull/91 ✅

Relies on https://github.com/chromaui/chromatic/pull/7724 ✅ - ensure that after login the dialog doesn't auto close, instead it redirects you back to where you wanted to go (granting permission).

<details>
  <summary>📦 Published PR as canary version: <code>0.0.86--canary.100.b09c8e2.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.86--canary.100.b09c8e2.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.86--canary.100.b09c8e2.0
  ```
</details>


src/utils/useChromaticDialog.ts
```diff
@@ -0,0 +1,53 @@
...
+    const handleMessage = ({ origin, data }: MessageEvent) => {
+      if (origin === dialogOrigin.current) {
+        const { message, ...rest } = data;
+        if (message in handlers) {
+          handlers[message as DialogEventName](rest);
+        }
+      }
+    };
+
+    window.addEventListener("message", handleMessage, false);
```
ghengeveld, line 26:
```suggestion
    window.addEventListener("message", handleMessage);
```
`false` is the default for `useCapture`. If you want to be explicit, it's better to use `{ capture: false }` which does the same.

src/utils/useChromaticDialog.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    return () => {
+      window.removeEventListener("message", handleMessage);
+    };
+  }, [handlers, dialogOrigin]);
```
ghengeveld, line 31:
```suggestion
  }, [handlers]);
```
Refs shouldn't be included in the dependency array, since they are guaranteed to be stable.

src/screens/Authentication/Verify.tsx
```diff
@@ -34,26 +35,9 @@ interface VerifyProps {
...
-      const options = `scrollbars=yes,width=${width},height=${height},top=${top},left=${left}`;
-      dialog.current = window.open(verificationUrl, "oauth-dialog", options);
-      if (window.focus) dialog.current.focus();
-    } else {
-      dialog.current = window.open(verificationUrl, "_blank");
-    }
-  };
+  const openChromatic = useChromaticDialog({
+    login: () => openChromatic(verificationUrl),
+  });
```
ghengeveld, line 40: Passing a plain object will cause the event listener in `useChromaticDialog` to be rebound on every render of `Verify`, since it's a fresh new `handlers` object each time. Probably better to wrap it in a `useMemo` first.

src/utils/useChromaticDialog.ts
```diff
@@ -0,0 +1,53 @@
...
+  // Close the dialog window when the screen gets unmounted.
+  useEffect(() => () => dialog.current?.close(), []);
+
+  useEffect(() => {
+    const handleMessage = ({ origin, data }: MessageEvent) => {
+      if (origin === dialogOrigin.current) {
+        const { message, ...rest } = data;
+        if (message in handlers) {
+          handlers[message as DialogEventName](rest);
+        }
```
ghengeveld, line 22: Maybe it's better if `useChromaticDialog` would take a single `messageHandler(message, rest)` that receives all messages, rather than doing the filtering in here. That would avoid having an object on the `useEffect` deps, which is really easy to accidentally recreate each render. The same is true of a function, but at least `useCallback` is more common than `useMemo`.

src/utils/useChromaticDialog.ts
```diff
@@ -0,0 +1,53 @@
...
+    (url: string) => {
+      const width = 800;
+      const height = 800;
+      const usePopup = window.innerWidth > width && window.innerHeight > height;
+
+      if (usePopup) {
+        const left = (window.innerWidth - width) / 2 + window.screenLeft;
+        const top = (window.innerHeight - height) / 2 + window.screenTop;
+        const options = `scrollbars=yes,width=${width},height=${height},top=${top},left=${left}`;
+        dialog.current = window.open(url, "oauth-dialog", options);
```
ghengeveld, line 43:
```suggestion
        dialog.current = window.open(url, "chromatic-dialog", options);
```
We could make this an option to `useChromaticDialog`.

ghengeveld (changes requested):

src/utils/useChromaticDialog.ts
tmeasday: I'll wait until we need it to add that.

src/utils/useChromaticDialog.ts
tmeasday: Hmm, but I guess there are typing advantages of passing an object. I'll see if I can make the typing work with a function `(eventName, eventData)`.

src/screens/Authentication/Verify.tsx
tmeasday: Good call.

src/utils/useChromaticDialog.ts
tmeasday: Figuring this out with @kasperpeulen. I'd like to get the pattern right.

kasperpeulen: You can make the typings work with a pattern like this:

image

https://www.typescriptlang.org/play?jsx=0#code/FAFwngDgpgBAogNygOxAZxgXhgbSk1AOQEMBbKALhgHIAjAG2OoBoYJix6B7YgEyrqN2nHr2oBdGAB8YufChAlyAgGZcuLNh258qAJnHBgKgK7IAxiACWXZDBNooAYQAWAJy6li18wBErxNwA5gAULsTIvPRQblQh8qhUiApoAJRYAHwwCFxWvOkA3sCywAC+Rua2aCAwDMRY9o6uHl4+-oFcoSF4BIpkUKzCOrzi6ZhZRbJWKjDxvUqwmEs0ddSFxbIwlchoXNEAdMEhQ6KpANwbsgD0V5t3dwB6APwbpTBQ9I4w07MJfeRYZbUNQadZ3ba7A5HE58c6XGA3e5I56vMrnIA

tmeasday:
Apologies @tevanoff this branch got messed up as I responded to feedback. I've fixed it up via merging in https://github.com/chromaui/addon-visual-tests/pull/91 and checked my QA instructions do work! Thanks.

ghengeveld (approved):

PR #104, opened 2023-09-15
Show a loading state in between URQL states

tmeasday:
To reproduce the problem, turn on network shaping to slow down your internet and switch stories. Rather than seeing the old story you should see a spinner while you wait:

image

Unfortunately there's no good way to simulate this behaviour in SB that I could come up with.

### DIFF
--- a/src/screens/VisualTests/VisualTests.tsx
+++ b/src/screens/VisualTests/VisualTests.tsx
@@ -59,7 +59,7 @@ export const VisualTests = ({
   // The user can choose when to change story (via sidebar) and build (via opting into new builds)
   const [storyBuildInfo, setStoryBuildInfo] = useState<StoryBuildInfo>({ storyId });
 
-  const [{ data, error }, rerun] = useQuery<
+  const [{ data, error, operation }, rerun] = useQuery<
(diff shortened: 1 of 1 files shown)


weeksling (approved):
👍 This also seems to fix the issue you called out on my pr with the baseline toggle not resetting after changing stories. This unmounts the component when data is loading and so it resets the state values to defaults.

src/screens/VisualTests/VisualTests.tsx
```diff
@@ -196,12 +201,12 @@ export const VisualTests = ({
...
 
-  return !storyBuild || error ? (
+  return !storyBuild || storyDataIsStale || error ? (
     <NoBuild
       {...{
         error,
-        hasData: !!data,
-        hasStoryBuild: !!storyBuild,
+        hasData: !!data && !storyDataIsStale,
+        hasStoryBuild: !storyBuild,
```
weeksling, line 209:
```suggestion
        hasStoryBuild: !!storyBuild,
```
Change needed to fix the no build state and stories

weeksling (approved):
One suggestion for broken stories, otherwise this seems good to go

PR #105, opened 2023-09-15
Clear access token on expiration or auth error

ghengeveld:
Clears the access token from memory and localStorage when either the token expires or the API returns either an auth error or no `viewer` data.

### DIFF
--- a/src/utils/graphQLClient.tsx
+++ b/src/utils/graphQLClient.tsx
@@ -1,30 +1,47 @@
 import { useAddonState } from "@storybook/manager-api";
+import { authExchange } from "@urql/exchange-auth";
 import React from "react";
-import { Client, fetchExchange, Provider } from "urql";
+import { Client, fetchExchange, mapExchange, Provider } from "urql";
 import { v4 as uuid } from "uuid";
 
 import { ACCESS_TOKEN_KEY, ADDON_ID, CHROMATIC_API_URL } from "../constants";
 
 export { Provider };
 
-let currentToken: string = localStorage.getItem(ACCESS_TOKEN_KEY);
-const accessTokenSharedStateKey = `${ADDON_ID}/accessToken`;
+let currentToken: string | undefined;
+let currentTokenExpiration: number | undefined;
+const setCurrentToken = (token: string | undefined) => {
+  try {
+    const { exp } = token ? JSON.parse(atob(token.split(".")[1])) : { exp: undefined };
+    currentToken = token;
+    currentTokenExpiration = exp;
+  } catch (e) {
+    currentToken = undefined;
+    currentTokenExpiration = undefined;
+  }
+  if (currentToken) {
+    localStorage.setItem(ACCESS_TOKEN_KEY, currentToken);
+  } else {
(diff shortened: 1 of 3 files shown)


src/utils/graphQLClient.tsx
```diff
@@ -33,12 +50,53 @@ const sessionId = uuid();
 
 export const client = new Client({
   url: CHROMATIC_API_URL,
-  exchanges: [fetchExchange], // no cacheExchange to prevent sharing data between stories
+  exchanges: [
+    // no cacheExchange to prevent sharing data between stories
```
tmeasday, line 54: @ghengeveld does this remove the need for this: https://linear.app/chromaui/issue/AP-3675/loading-state-when-switching-story?

src/utils/graphQLClient.tsx
```diff
@@ -33,12 +50,53 @@ const sessionId = uuid();
...
+    mapExchange({
+      onResult(result) {
+        if (result.data?.viewer === null) setCurrentToken(undefined);
```
tmeasday, line 57: Does this mean every query needs to include `Query.viewer`? What about mutations?

src/utils/graphQLClient.tsx
```diff
@@ -33,12 +50,53 @@ const sessionId = uuid();
...
+
+        // Prevent making a request if we know the token is missing, invalid or expired.
+        // This handler is called repeatedly so we avoid parsing the token each time.
+        willAuthError() {
+          if (!currentToken) return true;
+          try {
+            if (!currentTokenExpiration) {
+              const { exp } = JSON.parse(atob(currentToken.split(".")[1]));
+              currentTokenExpiration = exp;
+            }
```
tmeasday, line 86: Isn't this bit redundant? Don't we always set it above? At the least we should share this code.

tmeasday (commented):
Looks good although I have some questions. How to QA this?

src/utils/graphQLClient.tsx
ghengeveld: I don't think so. I didn't change anything in that regard (we didn't have caching before, and we still don't).

src/utils/graphQLClient.tsx
ghengeveld: If the query/mutation doesn't include a `viewer` it will be `undefined` which is not equal to `null`. That's what the `=== null` check is for.

src/utils/graphQLClient.tsx
ghengeveld: Ah yes, I forgot to clean this up after I made sure `currentTokenExpiration` is always set.

src/utils/graphQLClient.tsx
tmeasday: OK 👍

tmeasday: Maybe a comment for slower folks like me would help.

ghengeveld:
> Looks good although I have some questions. How to QA this?

@tmeasday I just manually messed up (edited) my JWT in localStorage.

tmeasday:
Nice!

tmeasday (approved):
LGTM, works well

PR #110, opened 2023-09-18
Turn on strict types (!)

tmeasday:
Mostly a pretty smooth process, a couple callouts.

<details>
  <summary>📦 Published PR as canary version: <code>0.0.79--canary.110.4594b11.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.79--canary.110.4594b11.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.79--canary.110.4594b11.0
  ```
</details>


src/components/SnapshotImage.tsx
```diff
@@ -65,10 +65,10 @@ export const Container = styled.div<{ href?: string; target?: string }>(
 );
 
 interface SnapshotImageProps {
-  componentName: Test["story"]["component"]["name"];
-  storyName: Test["story"]["name"];
+  componentName?: NonNullable<NonNullable<Test["story"]>["component"]>["name"];
```
tmeasday, line 68: @ghengeveld a lot of our API types are nullable when maybe they shouldn't be? I'm not sure if you have religion about this but it does make things a bit of a pain.

src/screens/LinkProject/LinkProject.tsx
```diff
@@ -123,7 +123,13 @@ function SelectProject({
   const [isSelectingProject, setSelectingProject] = useState(false);
 
   const handleSelectProject = React.useCallback(
-    (project: SelectProjectsQueryQuery["viewer"]["accounts"][number]["projects"][number]) => {
+    (
+      project: NonNullable<
+        NonNullable<
+          NonNullable<SelectProjectsQueryQuery["viewer"]>["accounts"][number]["projects"]
+        >[number]
+      >
```
tmeasday, line 131: @ghengeveld this one is particulary bad. I'm not sure why it's possible to have null results in the `accounts.projects` array.

src/utils/testsToStatusUpdate.ts
```diff
@@ -9,33 +9,42 @@ export const statusMap: Partial<Record<TestStatus, API_StatusValue>> = {
...
+      status && {
         status,
         title: "Visual Tests",
         description: "Chromatic Visual Tests",
       },
     ])
   );
 
+  // @ts-expect-error SB's API type is incorrect here you are allowed to pass null
   return update;
```
tmeasday, line 49: @ndelangen the type in SB's API is wrong. It says the values have to be `API_StatusObjects`:

https://github.com/storybookjs/storybook/blob/04ced719c25b81af8ab11813acf54f856b8fe3f8/code/lib/types/src/modules/api-stories.ts#L186

But then it does something different if they are `null`:

https://github.com/storybookjs/storybook/blob/04ced719c25b81af8ab11813acf54f856b8fe3f8/code/lib/manager-api/src/modules/stories.ts#L599

TBH I am not sure why TS isn't complaining about that line because if you look at the type of `value` it can never be `null` (according to the incorrect type definition):

image

Maybe because we don't have strict types on yet for `manager-api`? https://github.com/storybookjs/storybook/blob/04ced719c25b81af8ab11813acf54f856b8fe3f8/code/lib/manager-api/tsconfig.json#L5

src/utils/playAll.ts
```diff
@@ -1,6 +1,6 @@
-import type { StoryObj } from "@storybook/react";
+import type { Meta, StoryObj } from "@storybook/react";
 
-export function playAll<Story extends StoryObj>(
+export function playAll<Story extends StoryObj<Meta<TCmpOrArgs>>, TCmpOrArgs>(
```
tmeasday, line 3: @kasperpeulen I wonder if you can help me here. The previous type definition seemed to "throw" away the args type specificity and then complain if you passed a specific story object into it (it would say that `args: {}` isn't assignable to `args: { something... }`)

This seems to fix it but feels like it's not really "correct", IDK?

src/screens/VisualTests/VisualTests.stories.tsx
```diff
@@ -150,7 +151,8 @@ const withBuilds = ({
   lastBuildOnBranch?: LastBuildOnBranchBuildFieldsFragment;
   userCanReview?: boolean;
 }) => {
-  return withGraphQLQueryResult(QueryBuild, {
+  // FIXME -- I'm not sure why I have to mess with the types in this way
+  return withGraphQLQueryResult<typeof QueryBuild, VariablesOf<typeof QueryBuild>>(QueryBuild, {
```
tmeasday, line 155: @kasperpeulen this feels like the same problem that I don't really understand with the `playAll` thing. The previous type for `withGraphQLQueryResult` wouldn't work because it seemed to "lose" the variables that are bound into a specific type instance of the `TypedDocumentNode`.

The above works but this type it doesn't seem to be able to infer it for us so it's very awkward to use. Maybe we can take a look together?

src/components/SnapshotImage.tsx
ghengeveld: The general thought is that fields that require a db lookup (or any async operation) should be nullable. However, since a test with no story/component info is rather useless, and there is no way to delete such data from the db (as would be possible for e.g. a user or a comment), I think it would be okay to mark `story` and `component` non nullable. I'm happy if you make that change.

src/screens/LinkProject/LinkProject.tsx
ghengeveld: This gets into permissions territory. `null` is also returned if you don't have access to a resource, so as not to reveal its existence. However, it doesn't make sense to include null values in an array, so that seems wrong.

src/utils/playAll.ts
kasperpeulen: I can imagine that the StoryObj doesn't really work well like a generic constraint, as it is so complex with its conditional types. Maybe it is better to use `StoryAnnotations` which has no conditions:

```ts
import { ReactRenderer } from "@storybook/react";
import { StoryAnnotations } from "@storybook/types";

export function playAll<Story extends StoryAnnotations<ReactRenderer>>(
```

ghengeveld (approved):

src/utils/testsToStatusUpdate.ts
ndelangen: Indeed, not yet:
https://github.com/storybookjs/storybook/blob/702fd3e3ebf17e880fd82682a58b3f11e789f0c6/code/lib/manager-api/tsconfig.json#L5

src/screens/VisualTests/VisualTests.stories.tsx
kasperpeulen: I don't truly understand this issue, but just changing the function definition to this:

`function withGraphQLQueryResult<TQuery extends TypedDocumentNode<any, any>>(`

Make the inference in usages works properly again. Might be a TS bug.

src/utils/testsToStatusUpdate.ts
ndelangen: I opened a PR with a fix here:
https://github.com/storybookjs/storybook/pull/24206

src/screens/VisualTests/VisualTests.stories.tsx
tmeasday: OK, that's good enough for me; hardly the most crictical piece of code.

tmeasday: Thanks!

src/utils/playAll.ts
tmeasday: That works, great!

src/SidebarTop.tsx
```diff
@@ -41,6 +41,7 @@ export const SidebarTop = ({ api }: SidebarTopProps) => {
           name: "passed",
           color: color.positive,
         },
+        // @ts-expect-error SB needs a proper API for no link
         link: undefined,
```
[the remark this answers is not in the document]
tmeasday: Null doesn't work. IIRC `""` leads to it doing something a bit broken if you click on the notification.

src/utils/summarizeTests.ts
```diff
@@ -2,12 +2,12 @@ import { ComparisonResult, StoryTestFieldsFragment, TestResult, TestStatus } fro
...
-  if (statusCounts[TestStatus.Broken] > 0) return TestStatus.Broken;
-  if (statusCounts[TestStatus.Denied] > 0) return TestStatus.Denied;
-  if (statusCounts[TestStatus.Pending] > 0) return TestStatus.Pending;
-  if (statusCounts[TestStatus.Accepted] > 0) return TestStatus.Accepted;
+  if ((statusCounts[TestStatus.Failed] ?? 0) > 0) return TestStatus.Failed;
+  if ((statusCounts[TestStatus.InProgress] ?? 0) > 0) return TestStatus.InProgress;
+  if ((statusCounts[TestStatus.Broken] ?? 0) > 0) return TestStatus.Broken;
+  if ((statusCounts[TestStatus.Denied] ?? 0) > 0) return TestStatus.Denied;
+  if ((statusCounts[TestStatus.Pending] ?? 0) > 0) return TestStatus.Pending;
+  if ((statusCounts[TestStatus.Accepted] ?? 0) > 0) return TestStatus.Accepted;
```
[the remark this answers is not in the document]
tmeasday: True. You can even drop the `!!`. Do you think it's better, or more confusing that way?

src/components/SnapshotImage.tsx
tmeasday: I'll follow up with some non-null tweaks to our API.

ndelangen (approved):

src/utils/summarizeTests.ts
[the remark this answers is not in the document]
ghengeveld: Either works for me

PR #115, opened 2023-09-20
Setup TurboSnap

ghengeveld:
<details>
  <summary>📦 Published PR as canary version: <code>0.0.93--canary.115.a91e3de.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.93--canary.115.a91e3de.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.93--canary.115.a91e3de.0
  ```
</details>

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -15,4 +15,7 @@ jobs:
       - name: Publish to Chromatic
         uses: chromaui/action@v1
         with:
+          externals: |
+            - 'public/**'
+          onlyChanged: true
           projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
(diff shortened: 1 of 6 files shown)


ghengeveld:
> Why not put the settings in `x.config.json` and have it apply to the addon too?

@tmeasday What do you mean exactly? I can put `onlyChanged` and `externals` in `production.config.json` but it wouldn't work because the Action doesn't support the config file (at least it doesn't support the `configFile` option, maybe it does pick up the default?). 

You seem to suggest adding TurboSnap support to the addon itself too (so builds run through the addon can use TS)? That seems like a bigger project than what I'm trying to achieve here (speed up our CI builds).

tmeasday:
Builds through the addon will use turbo snap if it's configured in the config.json. I turned it on for the SB and CH monorepo and and haven't seen any issues yet. 

I guess I neglected to add support for the configFile option in the action, we should add that.

PR #116, opened 2023-09-20
When granting consent, immediately get access code rather than polling

tmeasday:
Using https://github.com/chromaui/chromatic/pull/7740, immediately on receiving the `grant` event, request a token from the server.

This partially solves AP-3549

<details>
  <summary>📦 Published PR as canary version: <code>0.0.95--canary.116.adc0530.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.95--canary.116.adc0530.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.95--canary.116.adc0530.0
  ```
</details>


tmeasday:
> This all looks pretty solid. Some tests?

Fair question. I'm not sure exactly what to test though.

Basically there are two functions in this file that put together REST requests in a slightly complicated way. Do we want to snapshot the data on those requests? I'm not quite sure there's a lot of utility in that (people aren't going to know if the snapshot is "right" or not, so will probably blindly accept changes). Really the test is sending the request to the server and seeing if it's happy.

The second one does some error handling logic also, but TBH I think we might revisit that as it's a bit weird.

ghengeveld:
I'm okay to leave the tests if there's no sensible strategy for it. Would be nice to have some E2E tests for this, eventually.

ghengeveld:
It seems the stories are failing because they use the manager-api which isn't available in the preview.

PR #124, opened 2023-10-03
Refactor Visual Tests stories to more closely control selected story

tmeasday:
This PR does 2 main things:

1. Refactor `VisualTests` so the `selectedBuildId` is a prop. This allows us to continue to write stories that vary the `selectedBuild` + `lastBuildOnBranch` without having weird stuff happen with the `selectedBuildId`. It also allows us to clearly assert on when the component should change the `selectedBuildId` in play functions[1].

2. Refactor the stories to use an args based approach and be very careful about the builds we pass. I'm pretty happy with how the stories end up reading but open to feedback.

[1] This also exposed a bug in how we are doing it right now that I'll fix in a follow up PR (https://github.com/chromaui/addon-visual-tests/pull/130)

---

Old text:

So the behaviour was actually right here, but the story didn't show it.

Let me explain because I think we might want to consider tightening up our testing strategy here.

 - We have a bunch of stories that hard code the returned `lastBuildOnBranch` and `selectedBuild`.
 - However, the `VisualTests` component [**changes** the `selectedBuildId` variable](https://github.com/chromaui/addon-visual-tests/blob/0030d1ef81fd19e57566fe83551d96c5a1648c09/src/utils/updateSelectedBuildInfo.ts#L6-L27) based on various criteria.
 - This was not reflected in the stories as the `selectedBuild` was hard-coded.

For the older story [`Empty Branch Local Build Captured Current Story`](https://www.chromatic.com/component?appId=6480e1b0042842f149cfd74c&csfId=screens-visualtests-visualtests--empty-branch-local-build-captured-current-story&buildNumber=1046&k=65145be13a8856438644adde-Bottom+2-up-up-false-true&h=9&b=-2) we fudge around this, because we set the `selectedBuild` to the `lastBuildOnBranch` when there is no `selectedBuild`:

https://github.com/chromaui/addon-visual-tests/blob/0030d1ef81fd19e57566fe83551d96c5a1648c09/src/screens/VisualTests/VisualTests.tsx#L158-L163

However, that doesn't help us for the newer, similar story [`Story Added In Last Build On Branch Not In Selected`](https://www.chromatic.com/component?appId=6480e1b0042842f149cfd74c&csfId=screens-visualtests-visualtests--story-added-in-last-build-on-branch-not-in-selected&buildNumber=1046&k=65145be13a8856438644add2-Bottom+2-up-up-false-false&h=21&b=-14) because the selected build *does* exist, it just doesn't have this story.

So the fix here is to make the GraphQL mock actually respond to the change in `selectedBuildId`. My question for y'all is: Should we, to be correct, make that change to all these stories? Or is this whole approach of testing the `VisualTests` component wrong and we should just write stories at the presentational level and do some other type of testing of "what happens when the `selectedStoryId` changes?" or "what happens when there's a new `lastBuildOnBranch`?"

Possibly we drop the quoted lines above too as it's sort of confusing there are two ways the `selectedBuild` ends up being equal to the `lastBuildOnBranch`. (Firstly by the short cut above, secondly by setting `selectedBuildId`, requerying, and re-rendering). Then again the shortcut saves the user waiting for a second query.

Another option would be to extend the logic to make `selectedBuild = lastBuildOnBranch` if the *current story* isn't in the actual selected build. Confusing?

<details>
  <summary>📦 Published PR as canary version: <code>0.0.106--canary.124.6cd88ba.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.106--canary.124.6cd88ba.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.106--canary.124.6cd88ba.0
  ```
</details>

### DIFF
--- a/src/screens/VisualTests/VisualTests.stories.tsx
+++ b/src/screens/VisualTests/VisualTests.stories.tsx
@@ -58,7 +58,7 @@ type LastOrSelectedBuildFragment = SelectedBuildFieldsFragment &
   LastBuildOnBranchBuildFieldsFragment;
 
 type QueryInput = {
-  /** If `lastBuildOnBranch` is unset, there will be no last build */
+  /** If `lastBuildOnBranch` is unset, there will be no last build on the branch */
   lastBuildOnBranch?: LastOrSelectedBuildFragment;
 
   /** If `selectedBuild` is unset, `lastBuildOnBranch` will be used *if* it matches `selectedBuildId` */
@@ -70,6 +70,10 @@ function mapQuery(
   { lastBuildOnBranch, selectedBuild: selectedBuildInput, userCanReview = true }: QueryInput,
   { selectedBuildId }: VariablesOf<typeof QueryBuild>
 ) {
+  if (selectedBuildInput && selectedBuildInput?.id !== selectedBuildId) {
+    throw new Error("Invalid story, selectedBuild does not match selectedBuildId");
+  }
+
   const selectedBuild =
     selectedBuildInput ??
     (lastBuildOnBranch?.id === selectedBuildId ? lastBuildOnBranch : undefined);
--- a/src/screens/VisualTests/VisualTests.stories.tsx
+++ b/src/screens/VisualTests/VisualTests.stories.tsx
@@ -566,29 +566,30 @@ export const PendingLocalBuildCapturedStory = {
       "https://www.figma.com/file/GFEbCgCVDtbZhngULbw2gP/Visual-testing-in-Storybook?type=design&node-id=2303-374529&t=qjmuGHxoALrVuhvX-0"
     ),
   },
-  play: async ({ canvasElement, args, argsByTarget }) => {
-    // We have to wait a moment for the story to render
-    const canvas = within(canvasElement);
-    await canvas.findAllByText("Latest");
+  // NOTE: this does not current work, the story is auto-selected
+  // play: async ({ canvasElement, args, argsByTarget }) => {
+  //   // We have to wait a moment for the story to render
+  //   const canvas = within(canvasElement);
+  //   await canvas.findAllByText("Latest");
 
-    // Ensure we don't switch to the new build, the user has to opt-in
-    mock(args.setSelectedBuildInfo!).mock.calls.forEach(([updater]) => {
-      const result = typeof updater === "function" ? updater(args.selectedBuildInfo) : updater;
-      expect(result).toEqual(args.selectedBuildInfo); // Unchanged
-    });
+  //   // Ensure we don't switch to the new build, the user has to opt-in
+  //   mock(args.setSelectedBuildInfo!).mock.calls.forEach(([updater]) => {
+  //     const result = typeof updater === "function" ? updater(args.selectedBuildInfo) : updater;
+  //     expect(result).toEqual(args.selectedBuildInfo); // Unchanged
+  //   });
 
-    // Now opt in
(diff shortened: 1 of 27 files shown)


src/screens/VisualTests/VisualTests.stories.tsx
```diff
@@ -87,24 +90,30 @@ const withGraphQLMutation = (...args: Parameters<typeof graphql.mutation>) => ({
...
-      name: "acme",
-      lastBuildOnBranch: lastBuildOnBranch || selectedBuild,
-    },
-    selectedBuild,
-    viewer: {
-      projectMembership: {
-        userCanReview,
+  return withGraphQLQueryResult(QueryBuild, ({ selectedBuildId }) => {
+    const selectedBuild =
+      typeof getSelectedBuild === "function" ? getSelectedBuild(selectedBuildId) : getSelectedBuild;
```
weeksling, line 104: I agree if there's a clear way to update all the stories to use the same 

Instead of every story passing in a function, since we know the behavior is to get the last build on branch by id, maybe we could just write the same function you passed in the new story here.
This way we only add the method once. The problem is we need to pass an ID on the lastBuildOnBranch in a story for it to work how we want.
```suggestion
      selectedBuildId === lastBuildOnBrach.id ? lastBuildOnBranch : selectedBuild
```

weeksling (approved):
Sorry for delayed review. This story makes sense to me. I'm not sure about changing every story to have a function but if the change could be handled in the withBuilds method then it makes sense.

Now that I understand them a bit better, I think a high level mock is useful. For edge cases and interactions might work well but I think this issue for instance might have been hard to test without a graphql mock that mimics our api even if it required changing. 

The story looks good to me. I can comment on the other questions mentioned tomorrow after a night's rest.

ghengeveld (approved):

ghengeveld:
> So the fix here is to make the GraphQL mock actually respond to the change in selectedBuildId. My question for y'all is: Should we, to be correct, make that change to all these stories? Or is this whole approach of testing the VisualTests component wrong and we should just write stories at the presentational level and do some other type of testing of "what happens when the selectedStoryId changes?" or "what happens when there's a new lastBuildOnBranch?"

I prefer the "integration test" style of stories over the pure presentational level, because it showcases what Storybook can do rather than copping out, while also providing us with better coverage. So I think the approach taken here is desirable. "some other type of testing" would likely mean E2E which I'd like to avoid if possible.

tmeasday:
Ok, so it sounds like the consensus here is to apply this approach to the other stories in the `VisualTests` component. I'll take another pass at it.

src/screens/VisualTests/VisualTests.stories.tsx
```diff
@@ -412,164 +461,190 @@ export const BrowserAddedInSelectedBuild: Story = {
...
+  //   await userEvent.click(viewLatestSnapshot);
+
+  //   const graphqlArgs = argsByTarget.graphql?.$graphql as typeof args.$graphql; // We need to type argsByTarget
+  //   await waitFor(() => {
+  //     expect(args.setSelectedBuildInfo).toHaveBeenCalledWith({
+  //       buildId: graphqlArgs?.AddonVisualTestsBuild?.lastBuildOnBranch?.id,
+  //       storyId: meta.args.storyId,
+  //     });
+  //   });
+  // },
```
tmeasday, line 600: As noted this play function is disabled here (and in 2 other tests) but fixed in https://github.com/chromaui/addon-visual-tests/pull/130

ghengeveld (approved):
Nice, I like this approach 👍

PR #126, opened 2023-10-03
git is required for visual test addon via chromatic cli

weeksling:
This adds an error state to the panel, disables the run button, and shows a notification if the addon is run in a project without git configured.

To test this:
 - Run storybook with addon installed.
 - Rename or delete .git directory to trigger an error.
 - It should show a single notification and update the panel with the git error.

Note: this currently stops the gitObservable loop from occurring and requires the server to be restarted after fixing the issue. Added language in the notification to prompt the user to restart storybook after fixing.

Note: This does not handle build errors once the gitInfo can be retrieved. The pr this branches off of should handle build-level errors including the git one commit issue.

<details>
  <summary>📦 Published PR as canary version: <code>0.0.99--canary.126.8f69032.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.99--canary.126.8f69032.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.99--canary.126.8f69032.0
  ```
</details>

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -41,7 +41,7 @@ const observeGitInfo = async (
       prev = gitInfo;
       timer = setTimeout(act, interval);
     } catch (e: any) {
-      console.error("getGitInfo failed. Ending loop.", e);
+      console.error(`Failed to fetch git info, with error:\n${e}`);
       errorCallback(e);
     }
   };
--- a/src/index.ts
+++ b/src/index.ts
@@ -27,10 +27,11 @@ function managerEntries(entry: string[] = []) {
 // Uses a recursive setTimeout instead of setInterval to avoid overlapping async calls.
 const observeGitInfo = async (
   interval: number,
-  callback: (info: GitInfo, prevInfo: GitInfo) => void,
+  callback: (info: GitInfo, prevInfo?: GitInfo) => void,
   errorCallback: (e: Error) => void
 ) => {
-  let prev: GitInfo;
+  let prev: GitInfo | undefined;
+  let prevError: Error | undefined;
   let timer: NodeJS.Timeout | undefined;
   const act = async () => {
     try {
@@ -39,10 +40,16 @@ const observeGitInfo = async (
         callback(gitInfo, prev);
       }
       prev = gitInfo;
+      prevError = undefined;
       timer = setTimeout(act, interval);
     } catch (e: any) {
-      console.error(`Failed to fetch git info, with error:\n${e}`);
-      errorCallback(e);
+      if (prevError?.message !== e.message) {
+        console.error(`Failed to fetch git info, with error:\n${e}`);
+        errorCallback(e);
+      }
+      prev = undefined;
+      prevError = e;
+      timer = setTimeout(act, interval);
     }
   };
   act();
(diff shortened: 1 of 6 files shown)


weeksling:
@MichaelArestad I added the notification here with the same title and subheader but added a final callout to restart the storybook server. Does that seem alright to you? 
Screen Shot 2023-10-03 at 7 42 34 PM

src/index.ts
```diff
@@ -26,17 +27,23 @@ function managerEntries(entry: string[] = []) {
...
-      callback(gitInfo, prev);
+    try {
+      const gitInfo = await getGitInfo();
+      if (Object.entries(gitInfo).some(([key, value]) => prev?.[key as keyof GitInfo] !== value)) {
+        callback(gitInfo, prev);
+      }
+      prev = gitInfo;
+      timer = setTimeout(act, interval);
+    } catch (e: any) {
+      console.error("getGitInfo failed. Ending loop.", e);
```
tmeasday, line 44:
```suggestion
      console.error(`Failed to fetch git info, with error:\n${e}`);
```

I'm not sure the stack trace is really all that helpful.

src/index.ts
```diff
@@ -99,9 +106,19 @@ async function serverChannel(
...
+
+  observeGitInfo(
+    5000,
+    (info) => {
+      gitInfoState.value = info;
+    },
+    (error: Error) => {
+      gitInfoError.value = error;
+    }
+  );
```
tmeasday, line 121: Theoretically, it's possible there's some other issue with git (like it gets corrupted or something). I'm not sure we really need to cope with that, but just flagging it here (a git failure doesn't necessarily mean git isn't setup).

tmeasday (approved):
LGTM and works well in testing.

I wonder if the notification would be weird for first time SB users who haven't even opened the addon yet. Perhaps we should play it cool and wait for them to reach the addon before showing them the message?

src/index.ts
weeksling: That makes sense. In this case, whatever the error occurring would otherwise fail to start the cli, so I think we'd want to handle it similarly to how we do. I think the stacktrace will be useful if users encounter other types of errors that cause this.

weeksling:
> I wonder if the notification would be weird for first time SB users who haven't even opened the addon yet. Perhaps we should play it cool and wait for them to reach the addon before showing them the message?

I agree. Removing the notification. I also updated it to keep the loop going so they don't have to restart storybook-server. The issue I thought I had with unsetting an addon state variable was an issue in my code

weeksling:
> > @MichaelArestad I added the notification here with the same title and subheader but added a final callout to restart the storybook server. Does that seem alright to you?
> 
> @weeksling In what scenario would folks see this? Is it necessary?

No I don't think this is necessary. I removed the notification entirely. Users will see the error when they try to check the addon panel. I thought this was necessary to draw attention to, but it will be happening most often for new storybooks so it's noisy.

MichaelArestad:
Perfect. Thank you!

src/index.ts
weeksling: Nevermind. This was another issue with my code.

PR #132, opened 2023-10-13
Handle unexpected errors from the runChromaticBuild method (cli misconfigured, missing build commmand, etc)

weeksling:
This is a quick fix to handle unexpected errors from the cli run method. The UI portion of the failure is handled by the onBuildProgressOrError, but the try-catch is necessary to prevent the server from crashing. Added a console.error to log the error that was caught, however this error is not as useful as the actual CLI logs. I felt it would be useful to keep in case an error occurs that the cli doesn't properly log for.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -109,7 +109,11 @@ async function serverChannel(
 
   channel.on(START_BUILD, async () => {
     const { projectToken } = projectInfoState.value || {};
-    await runChromaticBuild(localBuildProgress, { projectToken });
+    try {
+      await runChromaticBuild(localBuildProgress, { projectToken });
+    } catch (e) {
+      console.error(`Failed to run Chromatic build, with error:\n${e}`);
+    }
   });
 
   channel.on(STOP_BUILD, stopChromaticBuild);


src/index.ts
```diff
@@ -109,7 +109,11 @@ async function serverChannel(
 
   channel.on(START_BUILD, async () => {
     const { projectToken } = projectInfoState.value || {};
-    await runChromaticBuild(localBuildProgress, { projectToken });
+    try {
+      await runChromaticBuild(localBuildProgress, { projectToken });
+    } catch (e) {
+      console.error(`Failed to run Chromatic build, with error:\n${e}`);
```
weeksling, line 115: The log could be slightly redundant, but it does report different information than the cli does, and doesn't take up much additional space. I felt like it could be useful if the runChromaticBuild method were to fail for reasons the cli doesn't directly handle well.

If reviewer feels this is unnecessary/noisy I'm open to removing the extra log statement for now.
Example log of when the storybook-build command doesn't exist:
Screen Shot 2023-10-13 at 2 20 44 PM


```suggestion
  // Output handled by CLI
```

thafryer (approved):

PR #133, opened 2023-10-20
Replace `useAddonState` hook implementation to fix state sync bug

ghengeveld:
Manager <> server state were overriding each other, leading to the "run build" button to be stuck in "running" mode when refreshing the browser after running a build that broke.

I've reimplemented the shared state management tool as well as it's server API and manager hook. Besides cleaning things up, the main difference is that it now tracks an `index` property along with every state update, that will cause a state update to be rejected if the new index isn't higher than the current one. This way any race conditions will not cause weird behavior.

<details>
  <summary>📦 Published PR as canary version: <code>0.0.109--canary.133.29bf9a1.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.109--canary.133.29bf9a1.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.109--canary.133.29bf9a1.0
  ```
</details>


tmeasday (approved):
Good improvement!

So is the difference between `SharedState` and the old `/server` version mostly confined to addition of the index? It's a little hard to compare them but it seems like that's mostly it?

We might consider bringing this over to SB now it's a little more battle tested and you've implemented a manager/preview API (nicely, btw).

ghengeveld:
Thanks, I'm glad you like it.

Initially I had a different implementation in mind which is why I started a new SharedState concept rather than updating the original, but in the end I settled on basically the same behavior with the additional of the index. So yeah it's basically the same except for that index and the static constructor.

tmeasday:
WFM

PR #138, opened 2023-11-01
Refactor state management to useContext and useReducer

ghengeveld:
Telescoping on https://github.com/chromaui/addon-visual-tests/pull/141
Fixes AP-3815

Big refactor to get rid of a lot of prop drilling in favor of useContext. For the selectedBuild behavior I used a reducer.

I wouldn't say this is the final solution but it's making a step in the right direction.

<details>
  <summary>📦 Published PR as canary version: <code>0.0.116--canary.138.b91410d.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.116--canary.138.b91410d.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.116--canary.138.b91410d.0
  ```
</details>


src/screens/VisualTests/VisualTests.tsx
```diff
@@ -150,104 +93,229 @@ export const VisualTestsWithoutSelectedBuildId = ({
...
 
   // We always set status to the next build's status, as when we change to a new story we'll see
   // the next builds
-  const testsForStatus =
-    lastBuildOnBranch &&
-    "testsForStatus" in lastBuildOnBranch &&
-    lastBuildOnBranch.testsForStatus &&
-    getFragment(FragmentStatusTestFields, lastBuildOnBranch.testsForStatus.nodes);
+  useEffect(() => {
+    // if (!lastBuildOnBranchIsSelectable) return;
```
tmeasday, line 246: Why change this behaviour?

src/screens/VisualTests/VisualTests.tsx
```diff
@@ -150,104 +93,229 @@ export const VisualTestsWithoutSelectedBuildId = ({
...
+      ...testsToStatusUpdate(testsForStatus || []),
     }));
-    // We use the stringified version of buildStatusUpdate to do a deep diff
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [JSON.stringify(buildStatusUpdate), updateBuildStatus]);
 
-  const shouldSwitchToLastBuildOnBranch =
-    canSwitchToLastBuildOnBranch && lastBuildOnBranchCompletedStory;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [lastBuildOnBranchIsSelectable, JSON.stringify(lastBuildOnBranch), updateBuildStatus]);
```
tmeasday, line 261: Won't this send status unnecessary status updates if the `lastBuildOnBranch` changes but the status hasn't?

I would have thought this refactor shouldn't change behaviour.

src/utils/action.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import { action as act, type ActionOptions, type HandlerFunction } from "@storybook/addon-actions";
+import { jest } from "@storybook/jest";
+
+export const action = (name: string, options?: ActionOptions): HandlerFunction => {
+  const spy = jest.fn(act(name, options));
+  Object.defineProperty(spy, "name", { value: name, writable: false });
+  return spy;
+};
```
tmeasday, line 9: What's going on here? Doesn't SB do this for us?

src/screens/VisualTests/VisualTests.tsx
```diff
@@ -150,104 +93,229 @@ export const VisualTestsWithoutSelectedBuildId = ({
...
-      }}
-    />
+    <ReviewTestProvider watchState={reviewState}>
+      <SelectedBuildProvider watchState={selectedBuild}>
+        <BuildResults
+          {...{
+            branch: gitInfo.branch,
+            dismissBuildError,
+            localBuildProgress,
+            ...(lastBuildOnBranch && { lastBuildOnBranch }),
```
tmeasday, line 309: Isn't this available on the context now?

tmeasday (approved):
If you think this is better, then sure. I'm a bit worried because it was tricky to get all that selected build stuff right, and you've moved the logic for when we calculate it. But I guess if all the stories pass I'm OK with it.

I would suggest this is a PR that should get merged with 0 story changes overall.

src/utils/action.ts
ghengeveld: Only for args, not for parameters.

ghengeveld: I think this util should be part of `@storybook/jest` actually.

src/screens/VisualTests/VisualTests.tsx
ghengeveld: I didn't. This line wasn't supposed to be here anymore. We actually _do_ want to run the update even if the build isn't selectable (can't be switched to), because the update also clears the existing statuses and we still want that to happen. This is how it worked before too.

Effectively all I've done is move the `testsForStatus` and `testsToStatusUpdate` calculations inside the useEffect.

src/screens/VisualTests/VisualTests.tsx
ghengeveld: `lastBuildOnBranch` isn't, only `selectedBuild` is right now. I'll fix that though.

ghengeveld: Refactored `SelectedBuildContext` => `BuildContext` to store `lastBuildOnBranch` on there as well, along with the rest of `useBuild`'s data.

weeksling (changes requested):
I have some more review to do and I'll try to look more closely sometime this weekend.

But for now, I noticed this same bug in this pr. It seems like it resets the controls whenever the the polling reruns https://github.com/chromaui/addon-visual-tests/pull/139#pullrequestreview-1713568823

As well, I think we'll need the same context inside the onboarding and maybe other screens one day, so maybe we could add the ControlsContext and BuildContext to a separate directory, like src/state/ so we can easily find them and use them anywhere in the application if needed?

ghengeveld:
> I noticed this same bug in this pr. It seems like it resets the controls whenever the the polling reruns [#139 (review)](https://github.com/chromaui/addon-visual-tests/pull/139#pullrequestreview-1713568823)

Yes I fixed that on this PR just now. I'll update the other PR to have the latest from this one.

> As well, I think we'll need the same context inside the onboarding and maybe other screens one day, so maybe we could add the ControlsContext and BuildContext to a separate directory, like src/state/ so we can easily find them and use them anywhere in the application if needed?

Yeah I've been thinking about centralizing state in its own directory, though I'm not certain about it because in general I prefer to colocate and group by module/feature rather than by technical similarity. Controls should actually live higher up anyway I suppose (because it contains the warnings and settings toggles). For now I'll keep it as-is. We'll undoubtedly revisit it later when we build out more screens.

src/screens/VisualTests/VisualTests.tsx
tmeasday: That's not what I was concerned about. I was concerned about the change from the serialized update being the dependency to the serialized build.

src/screens/VisualTests/VisualTests.tsx
tmeasday: I still don't think you should change behaviour in a refactor PR.

src/utils/action.ts
tmeasday: Have you looked at the `@storybook/test` package?

src/screens/VisualTests/VisualTests.tsx
ghengeveld: Oh I see what you mean now. Hadn't considered that. I'll change it back.

src/screens/VisualTests/VisualTests.tsx
ghengeveld: I agree, and didn't intend to really change anything here. I hadn't considered the difference in behavior between checking the build vs the update.

src/screens/VisualTests/VisualTests.tsx
ghengeveld: Restored it now.

src/utils/action.ts
ghengeveld: Ah, I was looking in the wrong place.

ghengeveld: Ok so in the future we'll be able to just do `onSelectMode: fn()` (from `@storybook/test`) and it should all work, but this requires the SB alpha right now so I'm punting on that.

PR #139, opened 2023-11-01
Move switch and diff buttons

ghengeveld:
Telescoping on https://github.com/chromaui/addon-visual-tests/pull/138

<details>
  <summary>📦 Published PR as canary version: <code>0.0.117--canary.139.93bc623.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.117--canary.139.93bc623.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.117--canary.139.93bc623.0
  ```
</details>


weeksling:
Not specifically PR feedback in need of change, but just some general design feedback.

I find this orientation quite disorienting. I think I like the toggle on the bottom, but it feels like there is a lot of content on the top bar now. 
Screen Shot 2023-11-03 at 5 33 59 PM

That was without changes. With changes, it shifts around a bit because of the line break. Though I am really glad that it linebreaks because we can see the entire branch name. I wonder if the snapshot toggle could live on the bottom of the screen, above the browser and mode selectors, and then keep the top line for controls.  @MichaelArestad @ghengeveld 
Screen Shot 2023-11-03 at 5 38 36 PM

weeksling (changes requested):
I noticed a bug. I don't know if it's specific to this or the base pr, but if you toggle the diff image off, it switches back after a short pause. I don't know for certain, but I imagine this could be caused by polling since it always happens after a short delay, multiple times even.


https://github.com/chromaui/addon-visual-tests/assets/5882088/f69274f0-4c3f-4482-b3e4-8d718d7dac1b

ghengeveld:
@weeksling Thanks for the thoroughness. I fixed the diff toggle bug in the upstream PR. 

As for the design, I'm working on some improvements in another ticket. The switch button is going to move to the right next to the diff toggle. The accept buttons are moving up above those diff controls. 

I'm truncating (with ellipsis) the branch name. That's on just a single line though. Should we allow wrapping to a second line @MichaelArestad? That would be a bit more complicated (it's not a native CSS feature) but seems like a good idea.

MichaelArestad:
> Should we allow wrapping to a second line

No. Let's keep it to a single line for now for simplicity in both design and implementation.

ghengeveld:
Closing in favor of #148

PR #141, opened 2023-11-02
Fix warnings in browser console

ghengeveld:
Fixes AP-3814

- `:first-child` warning (even though irrelevant)
- `setState` on unmounted component
- `setState` in render (cannot update a component while rendering a different component)
- `<div> cannot appear as a descendant of <p>` in `BuildProgressInline`

This should also be a nice performance improvement, as we were needlessly setting up too many event listeners and had a bunch of crosstalk between them (I'm actually surprised it didn't make my browser really slow).

<details>
  <summary>📦 Published PR as canary version: <code>0.0.113--canary.141.66e051a.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.113--canary.141.66e051a.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.113--canary.141.66e051a.0
  ```
</details>

### DIFF
--- a/src/screens/VisualTests/StoryInfo.tsx
+++ b/src/screens/VisualTests/StoryInfo.tsx
@@ -37,7 +37,7 @@ const Info = styled.div(({ theme }) => ({
       fontSize: "inherit",
     },
 
-    "& > *:first-child": {
+    "& > span:first-of-type": {
       display: "inline-flex",
       alignItems: "center",
       height: 24,
@@ -133,12 +133,14 @@ export const StoryInfo = ({
   } else if (shouldSwitchToLastBuildOnBranch) {
     details = (
       <Info>
-        <b>
-          {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
-          <Link isButton onClick={switchToLastBuildOnBranch}>
-            View latest snapshot
-          </Link>
-        </b>
+        <span>
+          <b>
+            {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+            <Link isButton onClick={switchToLastBuildOnBranch}>
+              View latest snapshot
+            </Link>
+          </b>
+        </span>
         <span>Newer test results are available for this story</span>
       </Info>
     );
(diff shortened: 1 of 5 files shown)


src/screens/VisualTests/StoryInfo.tsx
```diff
@@ -133,12 +133,14 @@ export const StoryInfo = ({
...
-          </Link>
-        </b>
+        <span>
+          <b>
+            {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+            <Link isButton onClick={switchToLastBuildOnBranch}>
+              View latest snapshot
+            </Link>
+          </b>
+        </span>
```
weeksling, line 143: Was the console error that <b> could not be nested directly withing <Info>?

src/utils/SharedState.ts
```diff
@@ -35,14 +42,16 @@ export class SharedState<T = any> {
...
-      if (k !== key) return;
-      listeners.forEach((listener) => listener(v));
-    });
+    if (!instances.has(key)) {
+      instances.set(key, sharedState);
+      sharedState.channel.on(SET_VALUE, (k: string, v: T | undefined) => {
+        if (k !== key) return;
+        sharedState.listeners.forEach((listener) => listener(v));
+      });
+    }
```
weeksling, line 54: It looks like this should use https://github.com/chromaui/addon-visual-tests/pull/138 (refactor-state-management) as the base

weeksling (approved):
Seems like the base for this pr is off. Otherwise approving the code changes specific to this pr but reviewing the shared state work separately.

src/screens/VisualTests/StoryInfo.tsx
ghengeveld: No this was about the `:first-child` selector being vulnerable in SSR situations (hence not relevant, but a bad look nonetheless). I replaced it with `span:first-of-type` but that only works if they're all spans. This one wasn't and so I wrapped it.

src/utils/SharedState.ts
ghengeveld: What makes you think so? SharedState is a separate concept, unrelated to useContext which is applied in #138.

What I did here was avoid creating multiple instances of SharedState each time a component subscribes to it.

PR #143, opened 2023-11-06
Onboarding flow for creating first build

weeksling:
Adds an onboarding meant to run when a user creates a first build. The logic determining when to show this ui needs to be tweaked, but it builds the flow.

This UI, when shown unexpectedly, is weird because there could be a build but you can't access it because this is over top. This wouldn't be a problem if only used when there were no builds at all, but that might not end up being the case. So there may be other changes needed if we change when/how this is displayed.

This also serves as a base to build the guided onboarding on top of. 

https://github.com/chromaui/addon-visual-tests/assets/5882088/8df52869-c951-4932-9dfd-3352760e468c

Followup pr with proceeding guided tour to come.


weeksling:
Closing in favour of https://github.com/chromaui/addon-visual-tests/pull/144 which includes this and guided tour

PR #145, opened 2023-11-08
Authenticate with Chromatic CLI using OAuth token rather than `projectToken`

ghengeveld:
Fixes AP-3383

<details>
  <summary>📦 Published PR as canary version: <code>0.0.118--canary.145.0d30dd8.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.118--canary.145.0d30dd8.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.118--canary.145.0d30dd8.0
  ```
</details>


src/Panel.tsx
```diff
@@ -146,7 +146,7 @@ export const Panel = ({ active, api }: PanelProps) => {
         <VisualTests
           dismissBuildError={() => setLocalBuildProgress(undefined)}
           localBuildProgress={localBuildIsRightBranch ? localBuildProgress : undefined}
-          startDevBuild={() => emit(START_BUILD)}
```
tmeasday, line 149: Are we worried about emitting the access token over the channel unencrypted?

A malicious addon could watch for it easily enough. I think this is the full user token (not just a stripped down token that only has the ability to run builds ~= the project token), right?

tmeasday (dismissed):
We should also stop writing the project token to `chromatic.config.json` right?

src/Panel.tsx
ghengeveld: I asked this question [in Slack](https://chromaticqa.slack.com/archives/C02K88CKTFW/p1699269065837069). The token is already exposed to malicious addons, because it's in localStorage. Generally it's up to the user's own responsibility not to install untrusted addons.

And yes this is the user's OAuth token for the addon, not an app token (which is used to run builds).

tmeasday: I guess there is no way to encrypt it that couldn't be decrypted by an addon that has access to our source code.

ghengeveld: It's all client side, so no.

weeksling (approved):
👍

PR #146, opened 2023-11-09
Explain the `No Access` error and offer a way out

ghengeveld:
We already had a screen covering a situation where the user cannot access a project at all, but we didn't deal with the situation of insufficient permissions. This adds a screen similar to the existing one, specific to this scenario:

Screenshot 2023-11-09 at 11 39 36

<details>
  <summary>📦 Published PR as canary version: <code>0.0.115--canary.146.7b2009d.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.115--canary.146.7b2009d.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.115--canary.146.7b2009d.0
  ```
</details>

### DIFF
--- a/src/screens/VisualTests/NoBuild.tsx
+++ b/src/screens/VisualTests/NoBuild.tsx
@@ -88,11 +88,11 @@ export const NoBuild = ({
       return (
         <Container>
           <Heading>Network error</Heading>
-          <CenterText>{queryError.networkError}</CenterText>
+          <CenterText>{queryError.networkError.message}</CenterText>
           <br />
           {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
           <Link isButton onClick={() => setAccessToken(null)} withArrow>
-            Switch account
+            Log out
           </Link>
         </Container>
       );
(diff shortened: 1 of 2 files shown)


src/screens/VisualTests/NoBuild.tsx
```diff
@@ -84,13 +84,35 @@ export const NoBuild = ({
...
+          <Heading>{queryError.graphQLErrors[0].message}</Heading>
+          <CenterText>
+            {queryError.graphQLErrors[0].extensions.code === "FORBIDDEN"
+              ? "You may have insufficient permissions. Try logging out and back in again."
+              : "Try logging out or clear your browser's local storage."}
+          </CenterText>
+          <br />
+          {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+          <Link isButton onClick={() => setAccessToken(null)} withArrow>
+            Switch account
```
MichaelArestad, line 113: I might change this to "Log out" so a user doesn't think it's something different than it is or try to hunt for the more hidden logout button in the bottom toolbar.

thafryer (approved):
LGTM!

src/screens/VisualTests/NoBuild.tsx
ghengeveld: Ah, that's what I had at first and decided to keep it the same as the other screen. I agree though.

PR #153, opened 2023-11-29
Sort modes by their theme toolbar order

ghengeveld:
This attempts to order the list of modes in the ModeSelector to have the same ordering as the themes in the toolbar.

### DIFF
--- a/src/utils/useTests.ts
+++ b/src/utils/useTests.ts
@@ -1,4 +1,4 @@
-import { useGlobals } from "@storybook/manager-api";
+import { useGlobals, useGlobalTypes } from "@storybook/manager-api";
 import { useCallback, useState } from "react";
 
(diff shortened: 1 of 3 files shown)


src/utils/useTests.ts
```diff
@@ -19,7 +20,7 @@ const useGlobalValue = (key: string) => {
  * for the same story.
  */
 export function useTests(tests: StoryTestFieldsFragment[]) {
-  const theme = useGlobalValue("theme");
+  const [theme, themeConfig] = useGlobalValue("theme");
```
tmeasday, line 23:
```suggestion
  const [theme, themeType] = useGlobalValue("theme");
```

tmeasday (approved):
LGTM but I think the variable name is a touch confusing.

PR #154, opened 2023-12-11
Get CI green again.

tmeasday:
I accidentally pushed an update to SB@7.6.4 to main, which broke CI.

I'm not quite sure what the revert practice is so instead, we should maybe just fix it. This PR

1. Removes some `@ts-expect-errors` that were working around an SB bug that got fixed.
2. Approves changes to colors due to the `theme.textMutedColor` changing: https://github.com/storybookjs/storybook/pull/24525 (please confirm @MichaelArestad)
3. There's also [this broken story](https://www.chromatic.com/test?appId=6480e1b0042842f149cfd74c&id=6576517167d8d01812482c48), which was already problematic before the upgrade. Any idea what's happening here @ghengeveld ?


ghengeveld:
That story isn't broken. It intentionally doesn't show a baseline image because it's loading. We don't have an image loading indicator (yet).

ghengeveld (approved):

PR #158, opened 2024-01-24
UI to uninstall addon

JonathanKolnik:
image

image


JonathanKolnik:
We have to bump the dependency to sb 8, @ghengeveld can you help me do so

I assume it's these but just want to double check
```
 "peerDependencies": {
    "@storybook/blocks": "^7.2.0",
    "@storybook/client-logger": "^7.2.0",
    "@storybook/components": "^7.2.0",
    "@storybook/core-events": "^7.2.0",
    "@storybook/manager-api": "^7.2.0",
    "@storybook/preview-api": "^7.2.0",
    "@storybook/theming": "^7.2.0",
    "@storybook/types": "^7.2.0",
```

.storybook/main.ts
```diff
@@ -13,23 +13,18 @@ const configFileMap = {
...
-      options: {
-        configFile: configFileMap[CHROMATIC_BASE_URL || '"https://www.chromatic.com"'],
-      },
+  }, "@storybook/addon-interactions", "@storybook/addon-designs", {
+    name: useDistVersion ? "../dist/index.js" : "../src/dev.ts",
+    options: {
+      configFile: configFileMap[CHROMATIC_BASE_URL || '"https://www.chromatic.com"'],
     },
-  ],
+  }, "@storybook/addon-mdx-gfm"],
```
thafryer, line 27: Can we fix the formatting here?

thafryer (approved):
Looks good, barring there are no thoughts from the SB team on generating a full SB refresh.

shilman (changes requested):
The `removeAddon` feature is 8.0 only, so you should bump the versions here:

https://github.com/chromaui/addon-visual-tests/blob/main/package.json#L122-L129

Also, I don't think these should all be peer dependencies. AFAIK for strict package managers, peer dependency means that the package must be defined in the "parent" package (in this case the user's project). We don't want to require all of these packages in the user's `package.json`, so I think some (or all) of these should be moved to regular dependencies. Might need some discussion. I know @tmeasday has looked at this very recently.

tmeasday:
I'm not sure how regular SB packages are supposed to do it. Regular deps feels wrong because you might end up with the wrong versions.

tmeasday:
The versions are getting sorted in this PR: https://github.com/chromaui/addon-visual-tests/pull/159

PR #159, opened 2024-01-26
Rationalize SB deps

tmeasday:
We can't use peer deps because users won't have them in `package.json`. Also a lot of them were unused at this point.

<details>
  <summary>📦 Published PR as canary version: <code>0.0.125--canary.159.2e17e61.0</code></summary>
  <br />

  ```bash
  npm install @chromaui/addon-visual-tests@0.0.125--canary.159.2e17e61.0
  # or 
  yarn add @chromaui/addon-visual-tests@0.0.125--canary.159.2e17e61.0
  ```
</details>


package.json
```diff
@@ -52,12 +52,19 @@
     "typecheck": "tsc --noemit"
   },
   "dependencies": {
+    "@storybook/components": "^7.2.0 || ^8.0.0",
+    "@storybook/core-events": "^7.2.0 || ^8.0.0",
     "@storybook/design-system": "^7.15.15",
+    "@storybook/manager-api": "^7.2.0 || ^8.0.0",
+    "@storybook/theming": "^7.2.0 || ^8.0.0",
+    "@storybook/types": "^7.2.0 || ^8.0.0",
```
ndelangen, line 60: Considering you're using the new trick of:
```
import { globalPackages as globalManagerPackages } from "@storybook/manager/globals";
import { globalPackages as globalPreviewPackages } from "@storybook/preview/globals";
```

These storybook dependencies, should be devDependencies.

tmeasday:
I tested the canary with the Chromatic monorepo and it worked without issue.

tmeasday:
Thanks @JReinhold. @ndelangen do you think we should be building `src/index.js` twice (as this setup leads to [via this](https://github.com/chromaui/addon-visual-tests/blob/2e17e61fca3c9289f80ed6ff948dddd98b408a60/tsup.config.ts#L48)) given it is a preset and will only be used in `node`?

JReinhold:
> Thanks @JReinhold. @ndelangen do you think we should be building `src/index.js` twice (as this setup leads to [via this](https://github.com/chromaui/addon-visual-tests/blob/2e17e61fca3c9289f80ed6ff948dddd98b408a60/tsup.config.ts#L48)) given it is a preset and will only be used in `node`?

No I don't think that's necessary, in addon-onboarding we have [a fourth `nodeEntries` for this](https://github.com/storybookjs/addon-onboarding/blob/next/tsup.config.ts#L108-L117), which is the same as what we do throughout the monorepo as well.

I can't remember why it was left out of the addon-kit, might not have been intentional.

ndelangen (approved):

ndelangen:
@tmeasday some day in the future we'll want storybook to be able to consume ESM preset files.
(right now we don't)

So I'd argue you, can keep it like it is.

Storybook will for now just never use the `index.mjs`

tmeasday:
Ok will keep as is for now.

thafryer (approved):

PR #171, opened 2024-02-01
Replace the design system with local source files

yannbf:
Second attempt at fixing the issue, this time with a bit more harsh measure: copy the files used from `@storybook/design-system`, and get rid of the dependency altogether (for now at least)


src/components/design-system/Input.tsx
```diff
@@ -0,0 +1,443 @@
...
+  useCallback,
+  useRef,
+  useState,
+  forwardRef,
+  ReactNode,
+  ComponentProps,
+  MutableRefObject,
+} from 'react';
+import { styled, css } from '@storybook/theming';
+import { WithTooltip, TooltipMessage } from '@storybook/components';
```
yannbf, line 12: This was changed from 

```ts
import { WithTooltip, TooltipMessage } from '@storybook/design-system';
```

and I do not know whether there is a big difference, whether this will break or not. Please test it well

andrewortwein (approved):
Looks good 👍🏻

src/components/design-system/Input.tsx
kylegach: This may have had a minor impact: https://www.chromatic.com/review?appId=6480e1b0042842f149cfd74c&number=171&type=linked&view=changes

PR #172, opened 2024-02-02
QA Fast Follows

thafryer:
Fixes AP-4093, AP-4088, AP-4129, AP-4130, AP-4135, AP-4189

This PR seeks to resolve issues resulting from QA and feedback.

1. AP-4088, AP-4135, AP-4189 - At this point, the user is no longer required to set an origin. They still need to initialize the Git repo and commit at least one.
2. AP-4093 - If a user runs a project in a repo that does not have a chromatic script in their package.json, this interactive prompt will no longer appear in the CLI console.
3. AP-4129 - This adds a link to the user's Chromatic project in the footer menu of the addon. If there is no project set, we send them to https://www.chromatic.com/start, which will redirect them to their last visited account or to log in.
4. AP-4130, AP-4189 - This adds an option to remove the addon via the footer menu and from the "Git Not Detected" Screen.

<details>
  <summary>📦 Published PR as canary version: <code>1.0.5--canary.172.9d48858.0</code></summary>
  <br />

  ```bash
  npm install @chromatic-com/storybook@1.0.5--canary.172.9d48858.0
  # or 
  yarn add @chromatic-com/storybook@1.0.5--canary.172.9d48858.0
  ```
</details>

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -1,5 +1,7 @@
+import { action } from "@storybook/addon-actions"
 import { API, ManagerContext, State } from "@storybook/manager-api";
 import type { Decorator, Loader, Preview } from "@storybook/react";
+import { fn } from "@storybook/test"
 import {
   Global,
   ThemeProvider,
@@ -10,10 +12,12 @@ import {
   useTheme,
 } from "@storybook/theming";
 import { HttpResponse, graphql } from "msw";
+import { initialize, mswLoader } from "msw-storybook-addon";
 import React from "react";
 
 import { baseModes } from "../src/modes";
-import { initialize, mswLoader } from "msw-storybook-addon";
+import { UninstallProvider } from "../src/screens/Uninstalled/UninstallContext"
+
 
 // Initialize MSW
 initialize({
@@ -124,6 +128,14 @@ const withManagerApi: Decorator = (Story, { argsByTarget }) => (
   </ManagerContext.Provider>
 );
 
+const withUninstall: Decorator = (Story) => {
+  return (
+  <UninstallProvider>
+    <Story />
+  </UninstallProvider>
+  )
+}
+
 /**
  * An experiment with targeted args for GraphQL. This loader will serve a graphql
  * response for any arg nested under $graphql.
@@ -168,7 +180,7 @@ export const graphQLArgLoader: Loader = async ({ argTypes, argsByTarget, paramet
 };
 
 const preview: Preview = {
-  decorators: [withTheme, withManagerApi],
+  decorators: [withTheme, withUninstall, withManagerApi ],
   loaders: [graphQLArgLoader],
   parameters: {
     actions: {
@@ -195,6 +207,10 @@ const preview: Preview = {
   },
   argTypes: {
     $graphql: { target: "graphql" },
+    getChannel: { type: "function", target: "manager-api" },
+  },
+  args: {
+    getChannel: () => ({on: fn(), off: fn(), emit: action("channel.emit")}),
   },
   globalTypes: {
     theme: {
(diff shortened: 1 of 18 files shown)


.storybook/preview.tsx
```diff
@@ -195,6 +207,10 @@ const preview: Preview = {
   },
   argTypes: {
     $graphql: { target: "graphql" },
+    getChannel: { type: "function", target: "manager-api" },
+  },
+  args: {
+    getChannel: () => ({on: fn(), off: fn(), emit: action("channel.emit")}),
```
thafryer, line 213: When I use fn(), calls to emit do not show in the action tab. Also, notice that I have added `getChannel` as a global type if you try to write a play function and assert that getChannel.emit is called, you will run into some challenges.

src/Panel.tsx
```diff
@@ -60,15 +61,30 @@ export const Panel = ({ active, api }: PanelProps) => {
 
   // If the user creates a project in a dialog (either during login or later, it get set here)
   const [createdProjectId, setCreatedProjectId] = useState<Project["id"]>();
+  const [addonUninstalled, setAddonUninstalled] = useState<boolean>(false);
+
+  if (addonUninstalled) {
+    return (
+      <Provider key={PANEL_ID} value={client}>
+        <UninstallProvider>
```
tmeasday, line 69: Should we fold all these providers up into a single provider?

.storybook/preview.tsx
tmeasday: @kasperpeulen can you advise -- how best to use an action and assert?

tmeasday (approved):
LGTM code wise. I think the `Panel` component is getting a bit messy now, perhaps there's some simple things we can do to clean it up a bit.

tmeasday:
@thafryer 

> AP-4088, AP-4135, AP-4189 - At this point, the user is no longer required to set an origin. They still need to initialize the Git repo and commit at least one.
> AP-4093 - If a user runs a project in a repo that does not have a chromatic script in their package.json, this interactive prompt will no longer appear in the CLI console.

Are these two changes in the CLI? Are they part of this PR just via bumping the version.

In general, it would be better to have multiple PRs if different things are happening, IMO.

thafryer:
> @thafryer
> 
> > AP-4088, AP-4135, AP-4189 - At this point, the user is no longer required to set an origin. They still need to initialize the Git repo and commit at least one.
> > AP-4093 - If a user runs a project in a repo that does not have a chromatic script in their package.json, this interactive prompt will no longer appear in the CLI console.
> 
> Are these two changes in the CLI? Are they part of this PR just via bumping the version?
> 
> In general, it would be better to have multiple PRs if different things are happening, IMO.

Yes! AP-4088, AP-4135, and AP-4189 are resolved by bumping the CLI version and adjusting copy within the addon. AP-4093 is resolved by setting the interactive option to false when we kick off the builds.

These changes were small and where I started before adding the additional remove/uninstall workflows. I kept moving ahead since this PR was still pretty small with the noted changes..

src/Panel.tsx
thafryer: Yeah, I'll follow that up in subsequent PR.

.storybook/preview.tsx
thafryer: Yes, my preference is to use `fn`, and be able to assert in a play function and see that `emit` was called in the Actions tab.

tmeasday: @thafryer see this thread for more info: https://chromaticqa.slack.com/archives/C02K88CKTFW/p1706212067193149

kasperpeulen: The `fn` should already work and show to the action panel btw! @thafryer @tmeasday  
Sounds like a bug. Will check

PR #181, opened 2024-02-07
Fix kebab-case `box-shadow` to proper `boxShadow`

JReinhold:
There's currently an error in the console of any Storybook with VTA about using `box-shadow` while only `boxShadow` is allowed:

image

The kebab-cased style won't be applied so the warning is valid.

This was probably an oversight during the hectic transformation of CSS styles to JS object styles at https://github.com/chromaui/addon-visual-tests/commit/76466a58b444a16906317ce5e7493c3926cc4ad0#diff-6670cf4c6d5074d9ed37d761697fbbc372e5aa961f0907c7b305a1bd7f1e0e2aR118


thafryer (approved):
LGTM

JReinhold:
I just noticed that the `hoverEffect` is not used anywhere, so it could also be valid to remove it completely. Up to you.

PR #196, opened 2024-03-05
Show notification when connection to server is lost

ghengeveld:
Emits a heartbeat event from the server every second. Shows a notification in the browser if no heartbeat event is received for 3 seconds.

Screenshot 2024-03-05 at 15 23 15


thafryer (approved):

andrewortwein (approved):
📢 Nice touch! 👍🏻

andrewortwein (approved):
📢 Nice touch! 👍🏻

PR #197, opened 2024-03-05
Rerender Joyride when changing Storybook layout

ghengeveld:
Triggers a window resize event whenever the `layout` prop of Storybook's manager state changes. This [causes Joyride to recalculate](https://github.com/gilbarbara/react-joyride/discussions/630) its spotlight and tooltip positions.


thafryer (approved):
LGTM

PR #200, opened 2024-03-07
Clear all notifications when starting guided tour

ghengeveld:
This clears all notifications when the guided tour starts, so that there won't be a notification on top of the "𝑥 Changes" button.


thafryer (approved):
LGTM! This one is tough to crystalize in tests. 😅

PR #203, opened 2024-03-08
Preserve selected mode and/or browser, if possible, while navigating between stories

andrewortwein:
This allows you to select a particular mode/browser and, when you navigate to another story, show that same mode/browser (if it's available). Theoretically this is a nice way to quickly jump between stories.

Specifically, when navigating from one story to another, we will:
1. Try to load a test with changes that matches the same mode/browser
2. If that doesn't exist, try to load the first test with changes
3. If that doesn't exist, try to load an unchanged test that matches the same mode/browser
4. If that doesn't exist, load the first unchanged test

This puts an emphasis on tests with changes while still trying to preserve the mode/browser that the user selects.

Prior to this change, we were trying to load the mode that matches the theme selected in the SB toolbar. I removed this for now because, while nice, it felt incomplete to me. Modes can be more than just themes. If we selected a mode that matched _every_ SB global selected, that would feel consistent and expected. Until then, I prefer to not try to do it at all.

<details>
  <summary>📦 Published PR as canary version: <code>1.2.11--canary.203.d36a470.0</code></summary>
  <br />

  ```bash
  npm install @chromatic-com/storybook@1.2.11--canary.203.d36a470.0
  # or 
  yarn add @chromatic-com/storybook@1.2.11--canary.203.d36a470.0
  ```
</details>


thafryer (approved):
Looks Good! QA passed too!

PR #209, opened 2024-03-11
Wrap all screens with ControlsProvider and ignore config warnings on Authentication screens

ghengeveld:
Avoids showing config hints before they're relevant (before sign in). Also not all screens had the necessary ControlsProvider wrapper.


andrewortwein (approved):
:control_knobs: :+1:

PR #218, opened 2024-03-13
Only use criteria to select mode when story first loads

andrewortwein:
QA

1. Use a project with stories with modes:
   - One story should have a mode that is passed and then a mode that has changed, in that order
   - Another story should have the same modes, but both modes have changes
   - A third story should have additional modes that are not found in the first two stories, with changes
2. Navigate to the first story and verify that the second mode (the one with changes) is selected
3. Verify that you can change to the first mode and view the test
4. Navigate to the second story and verify that the same mode is shown from step 3
5. Navigate back to the first story and verify the the _unchanged_ mode is selected (this takes precedence over a selected mode that does not contain changes)
6. Navigate to the third story and select a mode that is unique to that story
7. Navigate back to the first story and verify that the mode with changes is selected

<details>
  <summary>📦 Published PR as canary version: <code>1.2.22--canary.218.31120f4.0</code></summary>
  <br />

  ```bash
  npm install @chromatic-com/storybook@1.2.22--canary.218.31120f4.0
  # or 
  yarn add @chromatic-com/storybook@1.2.22--canary.218.31120f4.0
  ```
</details>

### DIFF
--- a/src/screens/VisualTests/SnapshotComparison.stories.tsx
+++ b/src/screens/VisualTests/SnapshotComparison.stories.tsx
@@ -122,6 +122,13 @@ export const FirstPassed: Story = {
       })
     ),
   },
+  play: playAll(async ({ canvasElement, canvasIndex }) => {
+    const canvas = within(canvasElement);
+    const menu = await canvas.findByRole("button", { name: "Safari" });
+    await userEvent.click(menu);
+    const items = await screen.findAllByText("Chrome");
+    await userEvent.click(items[canvasIndex]);
+  }),
 } satisfies Story;
 
 export const StoryAdded: Story = {
--- a/src/screens/VisualTests/SnapshotComparison.stories.tsx
+++ b/src/screens/VisualTests/SnapshotComparison.stories.tsx
@@ -101,11 +101,9 @@ export const SpotlightOnly = {
 } satisfies Story;
 
 /**
- * Sort of confusing situation where the only comparison with changes (1200px/Safari) is on the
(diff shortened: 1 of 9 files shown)


ghengeveld (approved):

src/screens/VisualTests/SnapshotComparison.stories.tsx
```diff
@@ -122,6 +122,13 @@ export const FirstPassed: Story = {
       })
     ),
   },
+  play: playAll(async ({ canvasElement, canvasIndex }) => {
+    const canvas = within(canvasElement);
+    const menu = await canvas.findByRole("button", { name: "Safari" });
+    await userEvent.click(menu);
+    const items = await screen.findAllByText("Chrome");
+    await userEvent.click(items[canvasIndex]);
+  }),
```
andrewortwein, line 131: I added this `play` function since `useTests` now selects the first non-passed test/comparison by default. But, this story's comment kind of sounds like we expected to show the first passed test by default. Since that is no longer the case, is this story still valid at all?

andrewortwein:
@ghengeveld I made some updates to fix some `typecheck` problems and left a comment on a story I had to change. Please re-review this. Thanks.

src/screens/VisualTests/SnapshotComparison.stories.tsx
ghengeveld: Let's drop the `play` function, instead update the code comment above and update the snapshot.

ghengeveld (approved):

PR #219, opened 2024-03-13
VTA: Fix tooltip spacing

MichaelArestad:
All I am doing is removing a little margin from the tooltip.


thafryer (approved):
Code changes are fine!

ghengeveld (approved):
I would prefer keeping the name `TooltipNote` because that's what it's based on.

MichaelArestad:
> I would prefer keeping the name `TooltipNote` because that's what it's based on.

@ghengeveld Good call. I changed it to `TooltipNote`. This should be ready to go now once tests are done..

thafryer (approved):

PR #220, opened 2024-03-13
Set Zip to true by default on config updates

thafryer:
The Configuration Eyebrow shows on initial install. That's us telling the user to fix something out of the box that we can resolve for them. This does that.


andrewortwein (approved):
🤐👍🏻

PR #222, opened 2024-03-15
Normalize `configDir` before comparing

ghengeveld:
The configured `./ui/.storybook` was considered different than the suggested `ui/.storybook` value for `storybookConfigDir`, causing the addon to suggest an update. These values aren't actually different when normalized, so an update needn't be suggested. This fixes that by normalizing the suggested value before comparing to the configured value.


yannbf (approved):

andrewortwein (commented):
Good fix 👍🏻

PR #223, opened 2024-03-18
Fix dramatic layout bug during onboarding

MichaelArestad:
When we updated the Text component, our tests didn't catch this one. If you check `?path=/story/screens-onboarding--baseline-saved` on a wide viewport, you'll see it no longer wraps. The Text component was changed from `block` to `inline-block` allowing it to wrap in certain situations. I added a `boolean` `block` prop to Text allowing me to set it as a block. We could probably tackle this with Row, but Row came with a bunch of styles that would need to be overridden.

### DIFF
--- a/src/components/layout.tsx
+++ b/src/components/layout.tsx
@@ -1,7 +1,5 @@
 import { styled } from "@storybook/theming";
 
-import { Text as BaseText } from "./Text";
-
 export const Base = styled.div<{ hidden?: boolean }>(({ hidden, theme }) => ({
   background: theme.background.app,
   containerType: "size",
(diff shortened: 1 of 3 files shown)


thafryer (approved):
Looks good! For the tags, make sure that you use release instead of skip-release. This will ensure a new version is released

src/components/layout.tsx
```diff
@@ -1,7 +1,5 @@
 import { styled } from "@storybook/theming";
 
-import { Text as BaseText } from "./Text";
```
MichaelArestad, line 3: This was unused.

andrewortwein (approved):
🧱👍🏻

PR #226, opened 2024-03-19
Update guided tour tooltip to match Storybook

MichaelArestad:
This updates the tooltip to match Storybook's onboarding. It also simplifies the code a touch.
It would be nice if we had a story to cover this one, but I ran into issues trying. Would love someone to take this on when there is time for it.

To test, you may have to pull this down locally and add `&vtaOnboarding=true` after restarting Storybook.

It should look like this:
image

### DIFF
--- a/src/screens/GuidedTour/Tooltip.tsx
+++ b/src/screens/GuidedTour/Tooltip.tsx
@@ -19,14 +19,14 @@ const Wrapper = styled.div({
 });
 
 const TooltipTitle = styled.div(({ theme }) => ({
-  fontSize: 14,
+  fontSize: 13,
   lineHeight: "18px",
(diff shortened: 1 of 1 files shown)


src/screens/GuidedTour/Tooltip.tsx
```diff
@@ -40,10 +40,6 @@ const TooltipFooter = styled.div({
   marginTop: 15,
 });
 
-const NextButton = styled(Button)(
-  ({ secondary }) => secondary && { "&&:focus": { boxShadow: "none" } }
```
MichaelArestad, line 44: Not sure why this was here.

ghengeveld (approved):

PR #274, opened 2024-03-27
Save local state in `sessionStorage` or shared state

ghengeveld:
When switching to a docs page, addons don't render and thus the VTA addon is destroyed, losing all internal React state. By synchronizing local state with sessionStorage, this problem is circumvented.

<details>
  <summary>📦 Published PR as canary version: <code>1.3.0--canary.274.0843e22.0</code></summary>
  <br />

  ```bash
  npm install @chromatic-com/storybook@1.3.0--canary.274.0843e22.0
  # or 
  yarn add @chromatic-com/storybook@1.3.0--canary.274.0843e22.0
  ```
</details>


src/utils/useSessionState.ts
```diff
@@ -0,0 +1,79 @@
...
+    timeoutId = window.setTimeout(() => {
+      callback(...args); // Trailing edge call
+      timeoutId = 0;
+    }, wait);
+  };
+  debounced.cancel = cancel;
+  return debounced;
+};
+
+export function useSessionState<S>(
```
valentinpalkovic, line 24: This hook stores values into the sessionStorage AND has a local state in line 28 to preserve the current value. The problem with this approach is that if we have two components, which both use, let's say, `useSessionState('a,' initialState), and one of the components update the session state, the other one will not be notified since there aren't any session storage listeners in place. Indeed, my recommendation would be not to reinvent the wheel but to rely on existing solutions like `https://github.com/juliencrn/usehooks-ts/blob/master/packages/usehooks-ts/src/useSessionStorage/useSessionStorage.ts`

ghengeveld: Yeah, we should use `useSharedState` for that. I agree though that it's a footgun.
Generally I'm not keen on adding extra dependencies if I can avoid it. I'll take a look at this one though.

src/utils/useSessionState.ts
```diff
@@ -0,0 +1,105 @@
+import { Dispatch, SetStateAction, useCallback, useEffect, useMemo, useRef, useState } from "react";
```
valentinpalkovic, line 1: We should try to write automated tests were appropriate. I think a util function like this is pretty straightforward to test.

ghengeveld: Done

PR #300, opened 2024-05-08
Show message when snapshotting is disabled for story

ghengeveld:
Snapshots may be disabled using the `disable`, `disableSnapshot` or `docsOnly` parameter. If this happens, we now show a message explaining how to enable visual tests:

Screenshot 2024-05-08 at 12 43 45


JReinhold (approved):
Nice!!

PR #303, opened 2024-05-10
Poll for API connection and show notification when it fails

ghengeveld:
Runs a (server-side) GraphQL query every 5 seconds to check whether it's connected. If not, it prevents starting a new build:

Screenshot 2024-05-10 at 16 37 15

And shows a warning screen in the addon panel:

Screenshot 2024-05-13 at 14 16 01

These automatically go away when the connection is restored (delayed up to 5 seconds as it polls).

When the automatic retry is aborted after 10 attempts (~50 seconds), the user is prompted to retry manually:

Screenshot 2024-05-13 at 14 16 33

This way we don't just keep trying indefinitely when there is no internet connection.

<details>
  <summary>📦 Published PR as canary version: <code>1.4.0--canary.303.3053a3b.0</code></summary>
  <br />

  ```bash
  npm install @chromatic-com/storybook@1.4.0--canary.303.3053a3b.0
  # or 
  yarn add @chromatic-com/storybook@1.4.0--canary.303.3053a3b.0
  ```
</details>

### DIFF
--- a/src/Panel.tsx
+++ b/src/Panel.tsx
@@ -115,7 +115,7 @@ export const Panel = ({ active, api }: PanelProps) => {
   }
 
   if (apiInfo?.connected === false) {
-    return withProviders(<NoNetwork />);
+    return withProviders(<NoNetwork aborted={apiInfo.aborted} />);
   }
 
   // Render the Authentication flow if the user is not signed in.
--- a/src/Panel.tsx
+++ b/src/Panel.tsx
@@ -6,6 +6,7 @@ import { AuthProvider } from "./AuthContext";
 import { Spinner } from "./components/design-system";
 import {
   ADDON_ID,
+  API_INFO,
   GIT_INFO,
   GIT_INFO_ERROR,
   IS_OUTDATED,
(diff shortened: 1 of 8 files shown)


JReinhold:
Does this mean that SB will poll Chromatic every fifth second just by being open, regardless of if the addon is in active use or not?
And by active use I mean the addon's panel is in focus, or the user is waiting for VRT to finish, or similar.

I don't think that's the right trade-off:

1.  it's going to look super annoying in the DevTools Network panel
2. A notification will open whenever I loose connection to Chromatic, even if I'm not remotely working on that part of my Storybook. As a user I'd find that pretty annoying.
3. Paranoid users are going to hate this and call this idle tracking

Can't we just poll for network connection when the addon is actually in use? Like when the user is downloading diffs or waiting for results to finish?

ghengeveld:
@JReinhold I understand your concern and I think it makes sense to disable it when the addon hasn’t been interacted with for a while. But I’ll also address your specific concerns:

1. This request is server-side, so it won’t pollute the browser network logs. 
2. It’s pretty unlikely to not be able to connect to the API. It should probably only warn after 2 failed attempts though. 
3. Tracking seems a bit overblown since there’s not really any data being sent. Nevertheless I could be swayed to use a different/substitute internet connection check such as [is-online](https://www.npmjs.com/package/is-online). I do have plans to poll for some actual useful data (project config) as well though.

JReinhold:
Ah okay, sorry for my ignorance, that makes more sense when it's server side.

Then I'm fine with most of this actually, except I agree that the notification should be removed.

I understand that "polling when addon is in use" is then more complex, given you'd need to send "is panel open" data over the channel which might not be worth it.

PR #305, opened 2024-05-13
Allow interacting with Storybook UI through the Joyride overlay

ghengeveld:
The Joyride (guided tour) overlay normally prevents the user from interacting with the underlying page. However, it's possible for the "spotlight" area (which can be interacted with) to be positioned outside the viewport. When this happens it's impossible to continue or skip the tour since the entire screen is locked. By disabling user events on the overlay, it becomes possible to interact with the content below, including scrolling the sidebar. This is the simplest fix and while it's a little weird to be able to use Storybook with an overlay on top, I think that's a reasonable tradeoff.

https://github.com/chromaui/addon-visual-tests/assets/321738/d2153924-2ba7-40d9-803d-a053eb6f0308

It's a little weird that the overlay removes itself while scrolling, unfortunately I didn't find a way to prevent that.

### DIFF
--- a/src/screens/GuidedTour/GuidedTour.tsx
+++ b/src/screens/GuidedTour/GuidedTour.tsx
@@ -278,6 +278,7 @@ export const GuidedTour = ({
           overlay: {
             mixBlendMode: "unset",
             backgroundColor: "none",
+            pointerEvents: "none",
           },
           spotlight: {
             backgroundColor: "none",


valentinpalkovic:
Can you give some context in which particular scenarios the user must interact with the page? As I understood, it is just about scrolling to the spotlight area. Should we do this automatically instead? I am concerned that allowing to navigate on the page by the user will introduce more uncommon behavior. I would really love to prevent this and figure out a way for spotlights to always be in the viewport.

ghengeveld:
Closing in favor of #307

PR #309, opened 2024-05-16
Show spinner while snapshot image is loading

ghengeveld:
- Show a spinner while the snapshot image (latest/baseline, whichever is active) is loading.
- Prevent showing the diff/focus overlay image while the snapshot is loading.
- Show a spinner while the overlay image (diff/focus) is loading, on top of the snapshot image.

https://github.com/chromaui/addon-visual-tests/assets/321738/966a6f18-0bcf-4208-8bdb-f56dd4941ae3

<details>
  <summary>📦 Published PR as canary version: <code>1.5.0--canary.309.7b89f8d.0</code></summary>
  <br />

  ```bash
  npm install @chromatic-com/storybook@1.5.0--canary.309.7b89f8d.0
  # or 
  yarn add @chromatic-com/storybook@1.5.0--canary.309.7b89f8d.0
  ```
</details>


MichaelArestad:
Nice! I think the spinner is great.
1. Can you keep the spinner in place until all of the images are loaded? I'm not sure the pulsing image conveys it's not ready yet.
2. Can the spinner be vertically centered?

ghengeveld:
> Can you keep the spinner in place until all of the images are loaded? I'm not sure the pulsing image conveys it's not ready yet.

I could, but then we're not showing the user something useful while we could be. It's a micro optimization but I prefer that over showing the spinner for longer. In the Chromatic webapp we currently preload the snapshot image but we don't preload the overlay images, which inspired this behavior.

> Can the spinner be vertically centered?

It is, but the snapshot in the demo is < 100px tall and I've given the container a min-height: 100px, so it's actually getting vertically centered relative to that. Without the min-height the spinner could render flush to the top edge or even get clipped, which would be weird.

MichaelArestad:
> I could, but then we're not showing the user something useful while we could be. It's a micro optimization but I prefer that over showing the spinner for longer. In the Chromatic webapp we currently preload the snapshot image but we don't preload the overlay images, which inspired this behavior.

@ghengeveld In that case, I would recommend keep current behavior, but without the pulsing image.

> It is, but the snapshot in the demo is < 100px tall and I've given the container a min-height: 100px, so it's actually getting vertically centered relative to that. Without the min-height the spinner could render flush to the top edge or even get clipped, which would be weird.

Perfect!

ghengeveld:
@MichaelArestad Without the pulsating effect, there is no way to know the overlay image is still loading, which may lead people to incorrectly assume there are no visual changes.

An alternative effect I originally envisioned is a [linear indeterminate progress bar](https://m2.material.io/components/progress-indicators) either above or below the image. Do you think that would be more appropriate?

src/components/SnapshotImage.tsx
```diff
@@ -106,47 +141,87 @@ export const SnapshotImage = ({
...
+  const snapshotImageLoaded = baselineImageVisible ? baselineImageLoaded : latestImageLoaded;
+
+  let overlayImageLoaded = true;
+  if (showDiff && showFocus) {
+    overlayImageLoaded = comparisonImageLoaded && focusImageLoaded;
+  } else if (showDiff) {
+    overlayImageLoaded = comparisonImageLoaded;
+  } else if (showFocus) {
+    overlayImageLoaded = focusImageLoaded;
+  }
```
valentinpalkovic, line 157: I would prefer to use a `useMemo` or a simple `getOverlayImageLoaded` function here instead, which makes the code more readable and interrelated.

src/components/SnapshotImage.tsx
```diff
@@ -106,47 +141,87 @@ export const SnapshotImage = ({
...
-          src={latestImage.imageUrl}
+        <div
           style={{
-            display: baselineImageVisible ? "none" : "block",
+            position: baselineImageVisible ? "absolute" : "static",
+            visibility: baselineImageVisible ? "hidden" : "visible",
+            aspectRatio: `${latestImage.imageWidth} / ${latestImage.imageHeight}`,
+            width: latestImage.imageWidth,
+            maxWidth: "100%",
+            minHeight: 100,
```
valentinpalkovic, line 170: I prefer to reduce the JSX syntax to a minimum and outsource these inline stylings into a dedicated styled component so that the return part of this component is readable.

valentinpalkovic (approved):

PR #313, opened 2024-05-28
Add stories for `TooltipMenu` component

### LINKED ISSUE #240: TooltipMenu: create story for this component
From AP-4240

The TooltipMenu exists in code but there's no story for it which makes it tough for other developers to find or work on. 

Use case: I wanted to reuse TooltipMenu in another place but didn't see a story for it, which means I need to dig deeper into the 4 places it's reused to understand the API.

ghengeveld:
Fixes #240


thafryer (approved):
Looks Good!

PR #325, opened 2024-06-25
Use modern Yarn and run on port 6004

ghengeveld:
Make sure we use Yarn Berry rather than some ancient version.
Also, run Storybook on port 6004 to avoid port conflicts.


andrewortwein (changes requested):
Can you add a `packageManager` field to package.json? When I run an install, corepack keeps prompting me that one isn't there and that it will add it.

andrewortwein (approved):
🧶4️⃣👍🏻

PR #326, opened 2024-06-27
Replace API polling with browser-native "online" status check

ghengeveld:
Previously we would poll the Chromatic API every 5 seconds to check for service availability (and internet connection). This floods our GraphQL API with needless requests, so we decided to remove that polling behavior.

To cover the most likely scenario of no internet connection, the browser `online` and `offline` events are used to get notified of connection issues.

<details>
  <summary>📦 Published PR as canary version: <code>1.6.0--canary.326.58371ad.0</code></summary>
  <br />

  ```bash
  npm install @chromatic-com/storybook@1.6.0--canary.326.58371ad.0
  # or 
  yarn add @chromatic-com/storybook@1.6.0--canary.326.58371ad.0
  ```
</details>


src/preset.ts
```diff
@@ -123,14 +122,9 @@ const observeAPIInfo = (interval: number, callback: (apiInfo: APIInfoPayload) =>
...
     }
-    const ok = await fetch(CHROMATIC_API_URL, {
-      method: "POST",
-      headers: { "Content-Type": "application/json" },
-      body: JSON.stringify({ query: `{ viewer { id } }` }),
-    }).then(
-      (res) => res.ok,
-      () => false
-    );
+    const ok = await fetch("https://index.chromatic.com/healthz")
```
tevanoff, line 125: I'm likely missing something, can you explain why this is necessary? Is it not a rather safe assumption that there will be a stable connection on a dev machine, and if not everything will be broken? What do we gain from constantly checking this status rather than handling failures in the exceptional case that there is no connectivity?

ghengeveld: Yeah that's a fair point. We could just do without the service status check and rely fully on the browser's online/offline event. In the unlikely event that our API goes down, we could just handle that in the existing GraphQL requests (to show the service status message on a 503 response).

ghengeveld: I ripped it out already.

src/screens/NoNetwork/NoNetwork.tsx
```diff
@@ -1,59 +1,31 @@
...
-              Retry
-            </Button>
-          ) : (
-            <Button size="medium" variant="ghost" onClick={retry} disabled={retried}>
-              <SpinIcon />
-              Connecting...
-            </Button>
+
+          {!offline && (
+            <Link href="https://status.chromatic.com" target="_blank" rel="noreferrer" withArrow>
```
tevanoff, line 25: Nice!

tevanoff (approved):
I haven't QA'd but code looks good!

PR #331, opened 2024-07-16
Relay client-side fetch requests to the server using the Storybook channel API

### LINKED ISSUE #312: API requests should happen server-side and support a proxy agent
### Describe the bug

The browser network log is currently being flooded with Chromatic API calls as it polls for changes. This is annoying when attempting to work on a story that also makes network requests.

Furthermore, a HTTPS_PROXY isn't currently supported for these requests, even though the Chromatic CLI does support these.

### To reproduce

Just run the addon and look at the DevTools network tab.

### Environment

All versions.

ghengeveld:
Fixes #312 

This adds a `ChannelFetch` utility and `useChannelFetch` hook. ChannelFetch listens for messages on the `serverChannel`, performs `fetch` requests, and sends back a message with the response. The goal is relay client-side requests to the server, in order to avoid polluting the network request log in DevTools.

A `useChannelFetch` hook is included, which returns a `fetch` compatible function. This `fetch` replacement is passed to the URQL Client so all GraphQL requests will happen over the channel. Requests time out after 30 seconds (if they haven't already). The `signal` property is supported by means of a dedicated abort message.

<details>
  <summary>📦 Published PR as canary version: <code>1.9.0--canary.331.67b13f2.0</code></summary>
  <br />

  ```bash
  npm install @chromatic-com/storybook@1.9.0--canary.331.67b13f2.0
  # or 
  yarn add @chromatic-com/storybook@1.9.0--canary.331.67b13f2.0
  ```
</details>

### DIFF
--- a/src/utils/ChannelFetch.ts
+++ b/src/utils/ChannelFetch.ts
@@ -1,0 +1,48 @@
+import type { Channel } from "@storybook/channels";
+
+export const FETCH_ABORTED = "ChannelFetch/aborted";
+export const FETCH_REQUEST = "ChannelFetch/request";
+export const FETCH_RESPONSE = "ChannelFetch/response";
+
+type ChannelLike = Pick<Channel, "emit" | "on" | "off">;
+
+const instances = new Map<string, ChannelFetch>();
+
+export class ChannelFetch {
+  channel: ChannelLike;
+
+  abortControllers: Map<string, AbortController>;
+
+  constructor(channel: ChannelLike) {
+    this.channel = channel;
+    this.abortControllers = new Map<string, AbortController>();
+
+    this.channel.on(FETCH_ABORTED, ({ requestId }) => {
+      this.abortControllers.get(requestId)?.abort();
+      this.abortControllers.delete(requestId);
+    });
+
+    this.channel.on(FETCH_REQUEST, async ({ requestId, input, init }) => {
+      const controller = new AbortController();
+      this.abortControllers.set(requestId, controller);
+
+      try {
+        const res = await fetch(input as RequestInfo, { ...init, signal: controller.signal });
+        const body = await res.text();
+        const headers = Array.from(res.headers as any);
+        const response = { body, headers, status: res.status, statusText: res.statusText };
+        this.channel.emit(FETCH_RESPONSE, { requestId, response });
+      } catch (error) {
+        this.channel.emit(FETCH_RESPONSE, { requestId, error });
+      } finally {
+        this.abortControllers.delete(requestId);
+      }
+    });
+  }
+
+  static subscribe<T>(key: string, channel: ChannelLike) {
+    const instance = instances.get(key) || new ChannelFetch(channel);
+    if (!instances.has(key)) instances.set(key, instance);
+    return instance;
+  }
+}
--- a/src/utils/ChannelFetch.ts
+++ b/src/utils/ChannelFetch.ts
@@ -11,7 +11,7 @@ export class ChannelFetch {
 
   abortControllers: Map<string, AbortController>;
 
-  constructor(channel: ChannelLike) {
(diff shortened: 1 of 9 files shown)


src/utils/ChannelFetch.ts
```diff
@@ -0,0 +1,48 @@
+import type { Channel } from "@storybook/channels";
+
+export const FETCH_ABORTED = "ChannelFetch/aborted";
+export const FETCH_REQUEST = "ChannelFetch/request";
+export const FETCH_RESPONSE = "ChannelFetch/response";
```
ndelangen, line 5: Please namespace these events, as-is are a bit too generic.

A common practice is to prefix them with the addon's name/id.

src/utils/useChannelFetch.ts
```diff
@@ -0,0 +1,54 @@
+import { useChannel } from "@storybook/manager-api";
+
+const FETCH_ABORTED = "ChannelFetch/aborted";
+const FETCH_REQUEST = "ChannelFetch/request";
+const FETCH_RESPONSE = "ChannelFetch/response";
```
ndelangen, line 5: These events are duplicated, which could cause them to go out of sync.

A common practice is to have a `constants.ts` file, that's shared between manager/preview/server.

src/utils/useChannelFetch.ts
```diff
@@ -0,0 +1,54 @@
...
+        request.resolve(res);
+      }
+    },
+  });
+
+  return async (input: string | URL | Request, { signal, ...init }: RequestInit = {}) => {
+    const requestId = Math.random().toString(36).slice(2);
+    emit(FETCH_REQUEST, { requestId, input, init });
+
+    signal?.addEventListener("abort", () => emit(FETCH_ABORTED, { requestId }));
```
ndelangen, line 44: Maybe flip these statements around?

src/utils/graphQLClient.tsx
```diff
@@ -56,56 +54,69 @@ export const getFetchOptions = (token?: string) => ({
...
+    fetchOptions: getFetchOptions(), // Auth header (token) is handled by authExchange
+    ...options,
+  });
 
-export const GraphQLClientProvider = ({ children }: { children: React.ReactNode }) => {
-  return <Provider value={client}>{children}</Provider>;
+export const GraphQLClientProvider = ({
+  children,
+  value = createClient(),
+  ...rest
```
ndelangen, line 112: rest has no types

src/utils/ChannelFetch.ts
ghengeveld: I imagine this would be a generic Storybook utility, not something specific to this addon. I suppose I could prefix them until then.

src/utils/useChannelFetch.ts
ghengeveld: Yes these should be in Storybook core/events eventually.

src/utils/ChannelFetch.ts
ndelangen: Generic storybook APIs should not come from an addon.

This gives the impression it's a core feature, but then it's gone, if a users disabled/removed the addon.

ghengeveld: Ok, I namespaced them and moved to the addon constants file.

src/utils/graphQLClient.tsx
ghengeveld: I was doing this because of the `key` prop, but I’m not sure it’s necessary to pass that down.

ndelangen: You could consider using `FC` from `react for that.

src/utils/useChannelFetch.ts
ghengeveld: I guess mentally that makes sense. I'll also check for it to be aborted before emitting, we may not have to emit at all.

PR #334, opened 2024-08-07
Add `paramKey: "chromatic"` to allow disabling the VTA panel through story parameters

mellm0:
Configures the addon's `paramKey`, so that disabling this parameter will disable the Visual Tests panel in Storybook for a specific story:

```jsx
const MyStory = {
  parameters: {
    chromatic: {
      disable: true
    }
  }
}
```

Similarly, this can be set on `meta` instead to disable the VTA for all stories in the file.

The key is set to `chromatic` because this will cause Chromatic to skip the story when snapshotting. If you want to [disable snapshotting](https://www.chromatic.com/docs/ignoring-elements/#with-storybook) without disabling the addon panel, you can use the `disableSnapshot: true` parameter instead.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -27,6 +27,7 @@ export const SELECTED_BROWSER_ID = `${ADDON_ID}/selectedBrowserId`;
 export const TELEMETRY = `${ADDON_ID}/telemetry`;
 export const ENABLE_FILTER = `${ADDON_ID}/enableFilter`;
 export const REMOVE_ADDON = `${ADDON_ID}/removeAddon`;
+export const PARAM_KEY = ADDON_ID;
 
 export const CONFIG_OVERRIDES = {
   // Local changes should never be auto-accepted
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -27,7 +27,7 @@ export const SELECTED_BROWSER_ID = `${ADDON_ID}/selectedBrowserId`;
 export const TELEMETRY = `${ADDON_ID}/telemetry`;
 export const ENABLE_FILTER = `${ADDON_ID}/enableFilter`;
 export const REMOVE_ADDON = `${ADDON_ID}/removeAddon`;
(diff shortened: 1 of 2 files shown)


ghengeveld:
@mellm0 Thanks for the contribution! What is your intention here? The `paramKey` would allow you to disable/hide the addon panel (tab), but it wouldn't completely disable visual testing for a particular story. In fact if you'd disable the panel, the story would still be tested/snapshotted regardless. That's probably not what you expect?

If you'd expect snapshotting to be disabled for the story **and** have the panel disabled, then we could set `paramKey` to `"chromatic"`. Normally you'd [disable snapshotting for a story](https://www.chromatic.com/docs/ignoring-elements/#ignore-stories) using the `chromatic.disableSnapshot` parameter, but we actually also support `chromatic.disable` for historical reasons.

@tmeasday WDYT?

mellm0:
My intention is simply to have the ability to hide the panel on certain stories, though its true this does coincide with also disabling chromatic snapshots for it. 

I only chose the `ADDON_ID` as some other addons do the same, and I wanted it to be consistent. But I can change the name to `chromatic`, and that way it can hide at the same time. 

Might be a breaking change for users who expect to still see the panel, but are using `disable` instead of `disableSnapshot` to disable Chromatic snapshots. I wanted to avoid that too.

ghengeveld:
@mellm0 Sounds good. I think this is a legitimate use case so if you change it to "chromatic", I'm happy to merge this.

We're looking at updating some backend code to make `disable` behave identically to `disableSnapshot` as there's currently a slight difference, but as an end user it's not noticeable so it doesn't matter.

mellm0:
Updated 👍🏼

ghengeveld (approved):

ghengeveld:
Shipped in [v1.8.0](https://github.com/chromaui/addon-visual-tests/releases/tag/v1.8.0) 🚀

PR #342, opened 2024-10-11
Use new Test Provider API instead of SidebarTop and SidebarBottom

ghengeveld:
Updates the addon to use the new `experimental_TEST_PROVIDER` addon type that's coming in Storybook 8.4. When available, this API is used instead of the deprecated `experimental_SIDEBAR_TOP` and `experimental_SIDEBAR_BOTTOM` addon types.

The new API adds Visual tests as a test provider in the Testing Module at the bottom of the sidebar:

https://github.com/user-attachments/assets/76d83b87-e078-4132-98d3-2c53018f8335

Most notifications have been removed in favor of showing a warning text in the Testing Module.


src/manager.tsx
```diff
@@ -1,37 +1,220 @@
...
+          "change",
+          changedStoryCount.length
+        )}`
+      : "No visual changes detected";
+  }
+  return "Not run";
+};
+
 addons.register(ADDON_ID, (api) => {
+  if (global.CONFIG_TYPE !== "DEVELOPMENT") {
```
yannbf, line 188: 1. use globalThis instead
2. do you also want the panel to be disabled?
```suggestion
  if (globalThis.CONFIG_TYPE !== "DEVELOPMENT") {
```

ghengeveld: Yeah let's keep the panel, it has a special UI for static builds.

valentinpalkovic (approved):

PR #346, opened 2024-10-31
Fix dependencies to support Storybook >=`8.2.0` explicitly

JReinhold:
- Widened range of Storybook dependencies and peerDependencies to match `^8.0.0` as well as any current and future `8.X` prereleases
- `@storybook/channels` were only used for types, which we don't publish any of, so it became a `devDependency`
- `@storybook/types` were mainly used for types too, except for `Addon_TypesEnum` in the manager, but in the manager `@storybook/types` is globalized, so it can be a devDependency
- `@storybook/core-events` and `@storybook/telemetry` are properly used in the preset, so we have to depend on them.
- All the internally used Storybook's where changed from a canary to the 8.4.0 beta. I've vetted that they aren't actually used in the package but only in this repository's Storybook, except for the ones mentioned above.

<details>
  <summary>📦 Published PR as canary version: <code>3.2.2--canary.346.1e6e793.0</code></summary>
  <br />

  ```bash
  npm install @chromatic-com/storybook@3.2.2--canary.346.1e6e793.0
  # or 
  yarn add @chromatic-com/storybook@3.2.2--canary.346.1e6e793.0
  ```
</details>


package.json
```diff
@@ -53,9 +53,8 @@
     "jackspeak": "2.1.1"
   },
   "dependencies": {
-    "@storybook/channels": "^8.3.0",
-    "@storybook/telemetry": "^8.3.0",
-    "@storybook/types": "^8.3.0",
+    "@storybook/core-events": "^8.0.0 || ^8.1.0-0 || ^8.2.0-0 || ^8.3.0-0 || ^8.4.0-0 || ^8.5.0-0 || ^8.6.0-0",
+    "@storybook/telemetry": "^8.0.0 || ^8.1.0-0 || ^8.2.0-0 || ^8.3.0-0 || ^8.4.0-0 || ^8.5.0-0 || ^8.6.0-0",
```
vanessayuenn, line 57: Don't we want to gate it for 8.3+ because of the new Test Provider API?

ghengeveld, line 57: Just wondering about how this is going to resolve for this repository. Does the package manager pick the highest available option, or whichever is the first one in the pattern that's available? In the latter case I'm worried we might end up with v8.0.0 rather than a later minor version.

ghengeveld (approved):
Nice work!

JReinhold:
I've manually QA'd this:

- confirmed the canary works in `8.2`, `8.3` and `8.4` (fantastic to see the UI change over time here 👏)
- confirmed npm throws a peer dep error if you try this in an `8.1` project
- confirmed, as expected, that if you try `8.1.` anyway with `--legacy-peer-deps`, Storybook will throw a hard error.

ghengeveld:
:rocket: PR was released in [`v3.2.2`](https://github.com/chromaui/addon-visual-tests/releases/tag/v3.2.2) :rocket:

PR #349, opened 2024-12-19
Fix reading `status` of `undefined`

ghengeveld:
The `response` property may be `undefined`. This is not apparent from the URQL docs but checks out when looking at the type definitions.


yannbf (approved):

ghengeveld:
:rocket: PR was released in [`v3.2.3`](https://github.com/chromaui/addon-visual-tests/releases/tag/v3.2.3) :rocket:

PR #367, opened 2025-03-27
Migrate to new test provider API

JReinhold:


src/TestingModule.tsx
```diff
@@ -144,50 +187,86 @@ export const TestingModuleDescription = ({ api, running }: TestingModuleDescript
...
+    case localBuildProgress?.currentStep === 'complete':
+      description = localBuildProgress.errorCount
+        ? `Encountered ${pluralize('component error', localBuildProgress.errorCount, true)}`
+        : warningStatusCount
+          ? `Found ${pluralize('story', warningStatusCount, true)} with ${pluralize('change', warningStatusCount)}`
+          : 'No visual changes detected';
+      break;
+    default:
+      description = 'Not run';
+  }
```
JReinhold, line 215: this was essentially what the previous `TestingModuleDescription` component used to return. now we're setting it instead, and returning the JSX that `LegacyRender` would return - but simplified as it doesn't need to be generalized anymore.

src/constants.ts
```diff
@@ -1,6 +1,3 @@
-import * as coreEvents from 'storybook/internal/core-events';
-import { experimental_getStatusStore } from 'storybook/manager-api';
```
JReinhold, line 2: This was really bad, I don't know why I thought this was a good idea initially. Because the preset also imports from this `constants` file, it could result in the preset pulling in `manager-api` too.

src/Panel.tsx
```diff
@@ -1,17 +1,17 @@
...
 import { Spinner } from './components/design-system';
 import {
+  ADDON_ID,
   GIT_INFO,
   GIT_INFO_ERROR,
   IS_OFFLINE,
   IS_OUTDATED,
   LOCAL_BUILD_PROGRESS,
   REMOVE_ADDON,
-  statusStore,
```
JReinhold, line 14: see https://github.com/chromaui/addon-visual-tests/pull/367/files#r2017550766

package.json
```diff
@@ -109,7 +109,7 @@
     "react-dom": "^18.3.1",
     "react-joyride": "^2.7.2",
     "rimraf": "^3.0.2",
-    "storybook": "^9.0.0 || ^9.0.0-alpha.4",
+    "storybook": "0.0.0-pr-30875-sha-5d0152e4",
```
JReinhold, line 112: I will revert these versions back to 9.0.0 alphas before merging this PR, but we're waiting for a new alpha to release with the new TP API usage.

src/TestingModule.tsx
```diff
@@ -32,10 +32,44 @@ import { useBuildEvents } from './utils/useBuildEvents';
...
+
+const Progress = styled(ProgressSpinner)({
+  margin: 4,
+});
+
+const StopIcon = styled(StopAltIcon)({
+  width: 10,
+});
+
+export const TestingModule = () => {
```
ghengeveld, line 71: I would rename this to `TestProviderRender` for consistency with addon-test.

ghengeveld (approved):
Code-wise this looks fine. I'll do a functional pass as well.

.storybook/main.ts
```diff
@@ -14,16 +14,8 @@ const configFileMap = {
 
 const config: StorybookConfig = {
   addons: [
-    {
-      name: '@storybook/addon-essentials',
-      options: {
-        backgrounds: false,
-        viewport: false,
```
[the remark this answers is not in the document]
JReinhold: done, this now happens in `preview.tsx`

PR #399, opened 2025-12-20
Replace `filesize` dependency with internal util

FenPWysocki:
## Summary

Replace filesize with a tiny in-repo utility [formatBytesObject()](vscode-file://vscode-app/c:/Users/Piotr.Wysocki/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to format bytes and keep aligned units in upload progress.
## Changes
 - Added formatBytes.ts exporting formatBytesObject
- Removed filesize from package.json

## Behavior
Uses SI units (B, kB, MB, …) with base 1000 and 1-decimal rounding; preserves aligned-unit display (e.g., “1.2/3.4 MB”). No locale or bits support.
## Impact

 Smaller dependency footprint; tiny risk of minor formatting differences vs filesize on edge cases.
## Verification
 Local typecheck and build pass; no runtime logic changed beyond formatting. Optional: run tests (npm test).


ghengeveld:
@copilot resolve the merge conflicts in this pull request

src/utils/formatBytes.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  const computedExp = Math.min(UNITS.length - 1, Math.max(0, Math.floor(Math.log10(bytes) / 3)));
+  const exp = Math.min(UNITS.length - 1, Math.max(0, options?.exponent ?? computedExp));
+
+  const value = bytes / Math.pow(1000, exp);
+  const factor = Math.pow(10, round);
+  const rounded = Math.round(value * factor) / factor;
+
+  return { value: rounded, symbol: UNITS[exp], exponent: exp };
```
valentinpalkovic: **Unit-promotion regression vs `filesize`.**

`filesize` promotes the unit when rounding the value would display `1000` of the smaller unit. This util doesn't, so it diverges around common upload boundaries:

| bytes | old (`filesize`) | new |
|---|---|---|
| `999950` | `1 MB` | `1000 kB` |
| `999999` | `1 MB` | `1000 kB` |
| `999950000` | `1 GB` | `1000 MB` |

Upload progress label can render `1000/1000 kB` mid-upload instead of `1/1 MB`.

Fix after the rounding step:

```ts
let finalExp = exp;
let finalValue = rounded;
if (finalValue >= 1000 && finalExp < UNITS.length - 1 && options?.exponent === undefined) {
  finalExp += 1;
  finalValue = Math.round((bytes / Math.pow(1000, finalExp)) * factor) / factor;
}
return { value: finalValue, symbol: UNITS[finalExp], exponent: finalExp };
```

(Guard with `options?.exponent === undefined` so the shared-exponent call site keeps aligned units.)

src/utils/formatBytes.ts
```diff
@@ -0,0 +1,26 @@
+export type FormatBytesObject = {
+  value: number;
+  symbol: string;
+  exponent: number;
+};
+
+export function formatBytesObject(
```
valentinpalkovic, line 9: **Missing tests.** New util has no coverage, and this replaces a third-party lib whose semantics aren't exactly preserved.

Add `src/utils/formatBytes.test.ts` with boundary + edge cases:

- `999`, `1000`, `999949`, `999950`, `999999`, `1000000`
- explicit `exponent` option (shared-unit path used by `buildSteps.ts`)
- `0`, `NaN`, `Infinity`

Optional: JSDoc on the export noting SI base (1000) and that `exponent` forces unit alignment.

valentinpalkovic (commented):
Two realistic issues. Rest is noise.

1. **Unit-promotion regression** vs `filesize` — visible in upload labels near MB/GB boundaries (e.g. `999950` bytes → `1000 kB` instead of `1 MB`).
2. **Missing tests** for new util.

Details inline.

ghengeveld (approved):

PR #418, opened 2026-03-31
Add Share tool to quickly publish a Storybook

valentinpalkovic:
## Overview

Adds a sharing flow that lets a Storybook author publish their current local Storybook to a temporary, shareable Chromatic URL directly from the Storybook UI. The goal is to make it trivial to send a teammate, designer, or stakeholder a live link to the exact Storybook running on the author's machine — without asking them to clone, install, or run anything.

The flow lives behind a new "Share" button in the addon toolbar and walks the user through authentication, optional SAML SSO, subdomain selection, upload progress, and a final completion screen with a copyable URL. Shared links expire after a fixed window so they remain low-commitment.

## What's new

- **Share entry point** — A toolbar button in the addon panel opens a popover that hosts the entire share experience. Once a share is live, the popover surfaces the active link, who can access it, and when it expires.
- **Authentication via OAuth authorization code flow** — Replaces the previous device code flow. The browser-based grant is initiated from the panel, validated end-to-end with state checks, and the resulting code is exchanged for an access token. The same OAuth grant infrastructure now backs both the existing Verify screen and the new share flow.
- **Optional SAML SSO step** — When the account requires SSO, the flow detects this and prompts the user to sign in through their org's subdomain before continuing. SSO subdomains are also applied on staging hosts.
- **Subdomain selection** — Users pick or confirm a subdomain for the published Storybook. Input is normalized to lowercase and validated for hyphen rules so the resulting URL is always well-formed.
- **Publish progress and cancellation** — Upload progress is reported back to the panel as the build runs, and users can cancel mid-publish. Cancellation is matched against the originating request id so a stale cancel cannot terminate a newer publish.
- **Completion screen** — Shows the live URL with a one-click copy control, the access list, the publish timestamp, and the link expiry. A short-lived celebration banner appears immediately after a fresh publish, then collapses back into the standard timestamp/expiry row on subsequent popover opens.
- **Error recovery** — Failures surface a human-readable error and a retry control that restarts the flow cleanly on the first click. Unknown errors show a generic message instead of an empty string.
- **State machine** — The share flow is implemented as a `useReducer`-driven state machine with explicit terminal states, idempotent transitions, and guards against impossible re-entries (welcome skip, terminal progress, remount races).
- **"Outdated" behavior** — Now derived from Git information rather than ad-hoc heuristics, so the panel correctly reflects whether the current working tree matches what was last published.

## Security and robustness

- OAuth state is generated and validated on every grant outcome, including error paths.
- The token exchange endpoint is resolved from the same host as the grant, preventing host mismatch.
- The popup callback only posts `code` or `error` — the unreachable `denied` outcome and other dead code paths have been removed.
- Notification gating ensures the "share complete" toast cannot fire after cancellation or remount.

## Tests

New unit and integration coverage for:

- The OAuth grant parser and access token request
- The `useChromaticDialog` popup hook
- The full share flow rendered in the panel (idle, welcome, subdomain, SSO, uploading, complete, error)
- The preset-side wiring that drives publish progress and completion notifications
- The manager-side wiring (test environment was also updated to align with the secure error gating)

### Happy path: first-time share

1. Click the **Share** button in the toolbar. The popover should open on the welcome screen.
2. Click through to begin authentication. A browser window/tab should open for OAuth sign-in.
3. Approve the grant. The popover should advance past the auth step automatically once the callback fires.
4. Watch the **Uploading** screen. A shareable URL should be available shortly.
5. When the upload finishes, the popover should switch to the **Complete** screen showing:
   - A celebration banner ("Storybook published!") for roughly five seconds.
   - The shareable URL with a copy control.
   - The "Who has access" row.
   - A timestamp and an expiry message.
6. Click the copy control and paste the URL elsewhere — it should be the live Chromatic URL.
7. Open the pasted URL in a private window. The Storybook should load and match what's running locally.

### SAML SSO path

1. Sign in with an SSO-required account.
2. After the OAuth grant, the flow should pause on an **SSO** step prompting for the org subdomain.
3. Provide the subdomain and complete the SSO sign-in. You can find SSO testing instructions [here](https://www.notion.so/chromatic-ui/Testing-SAML-SSO-Okta-1626e81620348006bd9ec56710db9cc3)
4. Confirm the rest of the flow proceeds as in the happy path.
5. Repeat against a staging host and confirm the subdomain is applied there too.

### Cancellation

1. Start a share. While on the **Uploading** screen, click cancel.
2. Confirm:
   - The flow returns to a non-running state.
   - No "share complete" notification fires.
   - Starting a new share immediately afterwards works and is not affected by the cancelled run.

### Re-opening the popover after a successful share

1. After a successful publish, close the popover.
2. Wait more than five seconds, then reopen it.
3. The celebration banner should be gone; only the timestamp/expiry row should be shown alongside the URL and access list.
4. Reopening within five seconds of publish should still show the celebration.

### "Outdated" indicator

1. Publish a share, then make a local commit (or modify tracked files).
2. Reopen the popover. The state should reflect that the local working tree is now ahead of the published share.
3. Reset the working tree to the published commit and confirm the indicator clears.

### Login state synced between VTA Panel and Share Popup

1. When logging in in VTA, a login in the share popup should take place as well, and vice versa
2. Logging out in VTA should also log you out in the share pop-up

### Regressions to watch for

- Existing **Verify** flow should still authenticate the user normally — it now shares OAuth grant parsing with the share flow.
- Running a normal Chromatic build (non-share) should be unaffected.
- The toolbar's existing buttons and the rest of the panel UI should render and behave as before.


ghengeveld:
I don't love the `ShareSection` name here, because it isn't really a section of anything. `ShareDialog` or `SharePopup` feels more accurate and aligns better with what this UI element actually is.

ghengeveld:
Two related structure questions:

1. Could `src/screens/ShareSection/popoverPresence.ts` be rolled into `src/screens/ShareSection/ShareSection.tsx`? It feels pretty tightly coupled to that component and may not need its own file. Also you're aliasing the import, might as well use more descriptive exports to begin with. I don't love "present" as a term here anyway, I'd call it "open" or "rendered".
2. Why do we have a barrel file for the ShareSection components (`src/screens/ShareSection/index.ts`) instead of importing directly from the concrete files? Curious what benefit we’re getting here.

ghengeveld (approved):

package.json
```diff
@@ -54,7 +54,7 @@
   },
   "dependencies": {
     "@neoconfetti/react": "^1.0.0",
-    "chromatic": "^13.3.4",
+    "chromatic": "16.0.1--canary.1262.23788470989.0",
```
ghengeveld: This needs to be a final version.

src/screens/ShareSection/WhoHasAccess.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <AccessRow>
+        <GlobeIcon
+          size={14}
+          color={theme.textMutedColor}
+          style={{ flexShrink: 0, borderRadius: 6 }}
+          aria-hidden="true"
+        />
+        <AccessText>Anyone with the link</AccessText>
+        <LearnMoreLink
+          href="https://www.chromatic.com/docs/sharing"
```
kylegach:
```suggestion
          href="https://storybook.js.org/docs/sharing#quick-sharing"
```

PR #420, opened 2026-04-09
Clear story statuses when `onClearAll` is called by the Storybook manager

ghengeveld:
In preparation for https://github.com/storybookjs/storybook/pull/34478

As Storybook's UI will no longer actively clear story statuses when the "sweep" button in the testing module is used, VTA will have to listen for this event and clear the statuses itself.


ghengeveld:
Closed in favor of [421](https://github.com/chromaui/addon-visual-tests/pull/421)

PR #422, opened 2026-04-14
fix: move Chromatic CLI to devDeps (to avoid duplicated install)

theoludwig:
The `chromatic` dependency is only used as a CLI, not used in the package code itself (unless I'm wrong/missed something).

Moreover, it's outdated, currently set as v13.3.4, but latest is v16.3.0.
So, because I use a newer version in my project, the CLI is installed twice unnecessarily (detected thanks to [Node Modules Inspector](https://node-modules.dev/)):

image

One could argue, that it's "useless" (or not relevant) to install `@chromatic-com/storybook` without the Chromatic CLI in most cases, so we could put it as a peer dep instead, but that doesn't really help in anything.

I believe the best way to solve this "duplicated installation" issue, it's to put it in dev deps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,6 @@
   },
   "dependencies": {
     "@neoconfetti/react": "^1.0.0",
-    "chromatic": "^13.3.4",
     "filesize": "^10.0.12",
     "jsonfile": "^6.1.0",
     "strip-ansi": "^7.1.0"
@@ -82,6 +81,7 @@
     "@vitest/coverage-v8": "^3.0.8",
     "auto": "^11.0.5",
     "boxen": "^5.0.1",
+    "chromatic": "^13.3.4",
     "date-fns": "^2.30.0",
     "dedent": "^0.7.0",
     "eslint": "^9.21.0",


theoludwig:
cc @ghengeveld

ghengeveld:
Hi @theoludwig,

The chromatic-cli is listed as a dependency because we're [using its Node API](https://github.com/chromaui/addon-visual-tests/blob/d7bf2773630b15b9bf3cfc730233cc0149f81194/src/runChromaticBuild.ts#L1). This also means we are version-locked into its API.

Moving it to a devDependency won't fix the duplicate dependency issue, only a peerDependency could resolve that. However, we do not want to require users to install the chromatic-cli manually, since most users don't actually have it installed yet. Our general recommendation is to use `npx chromatic` which first installs the CLI if it isn't already. You could adopt that strategy if you want to avoid the duplicate dependency version.

I'll upgrade the chromatic-cli dependency to the most recent version, because it is quite outdated at this point.

theoludwig:
Thanks for the reply.

> The chromatic-cli is listed as a dependency because we're [using its Node API](https://github.com/chromaui/addon-visual-tests/blob/d7bf2773630b15b9bf3cfc730233cc0149f81194/src/runChromaticBuild.ts#L1). This also means we are version-locked into its API.

Oh okay, I have searched for `chromatic` not  `chromatic/node`, that's why I found nothing.

> I'll upgrade the chromatic-cli dependency to the most recent version

That would be nice. :+1:

PR #423, opened 2026-04-16
Fix VTA SSO authorize host for tenant subdomains

ethriel3695:
## Summary
- route the VTA SSO OAuth authorize request through the tenant subdomain instead of always using `www.chromatic.com`
- reuse the same subdomain rewrite for the verification URL so the SSO device flow stays on the same tenant host
- add a regression test covering both the SSO and default sign-in paths

## How to QA
1. Run `yarn vitest run src/utils/requestAccessToken.test.ts --reporter=verbose`.
2. In Storybook while signed out of the Visual Tests addon, open the panel and click `Sign in with SSO`.
3. Enter an SSO tenant subdomain such as `storybook` and continue.
4. Complete the SSO flow in the popup and confirm the addon proceeds to project selection or the linked-project state instead of showing a failed authentication/projects error.
5. Repeat the standard `Sign in with Chromatic` flow and confirm the non-SSO path still uses the default host successfully.


ethriel3695:
@copilot resolve the merge conflicts in this pull request

ghengeveld:
Fixed in #418

PR #425, opened 2026-05-05
Add refresh-token auth flow and harden HTML injection paths

ghengeveld:
## Summary
- add refresh-token based auth renewal in the manager client using a validated auth blob (`accessToken`, `refreshToken`, `tokenEndpoint`, `sessionId`) stored under the existing storage key
- centralize reactive refresh in `authExchange` with in-tab single-flight behavior, 10s timeout, one retry for transient/5xx failures, and session clearing on terminal auth failure
- harden the UI against HTML/code injection by banning `dangerouslySetInnerHTML`/`innerHTML`-style sinks in ESLint and replacing `innerHTML` usage in `PulsatingEffect` with `textContent`

## Test plan
- [x] `yarn lint`
- [x] `yarn typecheck`
- [x] `yarn vitest run src/utils/graphQLClient.test.ts`

Made with [Cursor](https://cursor.com)


src/utils/graphQLClient.tsx
```diff
@@ -45,16 +102,80 @@ export const useAccessToken = () => {
...
+  const timeoutId = globalThis.setTimeout(() => abortController.abort(), REFRESH_TIMEOUT_MS);
+  try {
+    const nextAuth = await refreshAccessToken({
+      clientId: OAUTH_CLIENT_ID,
+      tokenEndpoint: currentAuth.tokenEndpoint,
+      refreshToken: currentAuth.refreshToken,
+      sessionId: currentAuth.sessionId,
+      signal: abortController.signal,
+    });
+    setCurrentAuth(nextAuth);
```
valentinpalkovic, line 189: **[High] Refreshed access token never reaches React state — builds run with stale token.**

`attemptTokenRefresh` updates the module-level `currentAuth` / `currentToken` and `localStorage`, but `useAccessToken` only updates via `setTokenState` inside `updateToken` (lines 94-100). The refresh path bypasses that callback entirely.

`useBuildEvents` reads `accessToken` from React state and emits it on `START_BUILD` (`src/utils/useBuildEvents.ts:11,36`). After a silent refresh, the React tree (and therefore `startBuild` → `preset.ts` `START_BUILD` handler) keeps the **expired** token. So a 401-triggered refresh succeeds for GraphQL but the next build still fires with the stale token.

Fix options:
- expose a small subscribe/notify on the module and have `useAccessToken` subscribe;
- read `currentToken` lazily inside `startBuild` instead of capturing via closure / props;
- push token change through `useAddonState` so `useAccessToken` re-renders.

Whichever you pick, add a regression test: simulate refresh → assert React-side `accessToken` updated.

_(Codex + Claude)_

src/utils/graphQLClient.tsx
```diff
@@ -45,16 +102,80 @@ export const useAccessToken = () => {
...
+    const statusMatch = error.message.match(/\((\d{3})\)/);
+    if (statusMatch) {
+      const statusCode = Number(statusMatch[1]);
+      return statusCode >= 500;
+    }
+    if (error.name === 'AbortError') {
+      return true;
+    }
+  }
+  return true;
```
valentinpalkovic: **[Medium] `isRetryableRefreshError` defaults to `true` for non-status-coded errors → retries terminal OAuth errors.**

```ts
if (error instanceof Error) {
  if (statusMatch) return statusCode >= 500;   // 4xx → false ✓
  if (error.name === 'AbortError') return true;
  // falls through ↓
}
return true;  // ← non-status Errors and non-Errors retry
```

Errors thrown by `decodeTokenResponse` (e.g. `invalid_grant`, `invalid_request`) carry no `(NNN)` token in the message. They fall through and retry once — wasted call against an already-revoked refresh token, plus extra latency before logout.

```ts
const isRetryableRefreshError = (error: unknown) => {
  if (error instanceof Error) {
    const statusMatch = error.message.match(/\((\d{3})\)/);
    if (statusMatch) {
      const statusCode = Number(statusMatch[1]);
      return statusCode >= 500;
    }
    if (error.name === 'AbortError') {
      return true;
    }
    return false;
  }
  return false;
};
```

_(Gemini + Claude)_

src/utils/graphQLClient.tsx
```diff
@@ -45,16 +102,80 @@ export const useAccessToken = () => {
...
+        await attemptTokenRefresh();
+      } catch (error) {
+        if (isRetryableRefreshError(error)) {
+          await attemptTokenRefresh();
+          return;
+        }
+        throw error;
+      }
+    })()
+      .catch(() => {
```
valentinpalkovic: **[High] Refresh failures are swallowed — `authExchange.refreshAuth` (line 204) sees success even when the session was cleared.**

The outer `.catch(() => { console.warn(...); clearCurrentAuth(); })` resolves the `refreshPromise` cleanly. urql's `authExchange.refreshAuth` awaits this and treats it as a successful refresh, then retries the original op (now without auth) instead of surfacing the auth error.

Fix: rethrow inside the catch (or remove the catch entirely and let it propagate, with `clearCurrentAuth()` in a `finally`). Pair with M7 (UX toast) so the user understands why they were just signed out.

```ts
.catch((error) => {
  clearCurrentAuth();
  throw error;
})
```

_(Codex + Claude)_

src/utils/graphQLClient.tsx
```diff
@@ -45,16 +102,80 @@ export const useAccessToken = () => {
...
+        console.warn('Session expired. Please sign in again.');
```
valentinpalkovic, line 205: **[Medium] No UI feedback on silent refresh failure.**

A `console.warn` is the only signal when a session expires server-side or the refresh token has been revoked. The user sees stale UI / failing requests until something else triggers an auth-gated render.

Suggest routing through `useErrorNotification` or emitting a session-expired toast on the addon channel when `clearCurrentAuth` is called from the refresh path (vs an explicit logout). That's also the natural pairing for the rethrow in H3.

_(Gemini)_

valentinpalkovic (commented):
## CCG review (Codex + Gemini + Claude synthesis)

Solid security upgrade overall. PKCE + auth-code is a clear win and the single-flight refresh pattern is the right shape. Findings cluster around (a) refreshed token never reaching React state, (b) refresh-error propagation, (c) `isRetryableRefreshError` defaulting to retry on terminal OAuth errors. No Critical. 3 High that should land before merge.

### Reviewer convergence

| Finding | Codex | Gemini | Claude |
|---|---|---|---|
| H1 Stale React token after refresh | ✅ | — | ✅ |
| H2 Premature `setAuthenticatedSession` | ✅ | — | ✅ |
| H3 Swallowed refresh error | ✅ | partial | ✅ |
| M1 `isRetryableRefreshError` default | — | ✅ | ✅ |
| M2 `refresh_token` required by schema | ✅ | — | ✅ |
| M3 Logout/refresh race | ✅ | — | ✅ |
| M5 No legacy storage migration | ✅ | — | partial |
| M6 Cross-tab refresh not single-flight | ✅ | — | ✅ |
| M7 Silent failure, no UX feedback | partial | ✅ | — |
| L2 Lint gaps (outerHTML, document.write, srcdoc) | ✅ | — | — |
| L3 `globalThis` vs `window` inconsistency | — | ✅ | ✅ |
| L4 `<React.Fragment />` regression | — | ✅ | — |

No conflicts between advisors.

### Action checklist (priority order)

1. **H3** — rethrow refresh errors so `authExchange.refreshAuth` sees failure.
2. **H1** — propagate refreshed access token to React state / `useAccessToken` subscribers.
3. **H2** — defer `setAuthenticatedSession` until project-creation flow completes.
4. **M1** — fix `isRetryableRefreshError` fallthrough.
5. **M3** — generation counter to discard refresh result after logout.
6. **M2** — make `refresh_token` optional in `TokenResponseSchema` (RFC 6749 §6).
7. **M7** — surface session-expired notification.
8. **M5** — document forced re-login on upgrade or migrate legacy values.
9. **M6** — decide cross-tab strategy (BroadcastChannel / `storage` event) or document.
10. **L2 / L3 / L4** — quick cleanups.
11. Tests for H1 / M3 / state-mismatch on error callback / single-flight concurrent 401 / refresh-without-rotation.

Further related findings outside this PR's diff (already in `next`): OAuth `state` not validated on the error sub-branch in `Verify.tsx`'s grant handler, and `state` forwarded with optional validation in `manager.tsx`. Per RFC 6749 §10.12 state must be validated on **all** authorization responses. Worth following up.

Detailed inline comments below.

src/utils/graphQLClient.tsx
```diff
@@ -1,31 +1,88 @@
...
     currentToken = null;
-    currentTokenExpiration = null;
   }
-  if (currentToken) {
-    localStorage.setItem(ACCESS_TOKEN_KEY, currentToken);
+  persistCurrentAuth();
+};
+
+const clearCurrentAuth = () => {
+  refreshAbortController?.abort();
```
valentinpalkovic, line 69: **[Medium] Logout ↔ in-flight refresh race can resurrect auth state.**

`clearCurrentAuth` aborts the controller and nulls module state, but if `attemptTokenRefresh`'s `fetch` already resolved (response in flight, abort no-ops), the awaited continuation still calls `setCurrentAuth(nextAuth)` (line 142) and re-populates auth post-logout.

Guard with a generation counter:

```ts
let authGen = 0;
const clearCurrentAuth = () => { authGen++; /* ...rest */ };
const attemptTokenRefresh = async () => {
  const myGen = authGen;
  const nextAuth = await refreshAccessToken({...});
  if (myGen !== authGen) return; // logged out mid-flight
  setCurrentAuth(nextAuth);
};
```

_(Codex + Claude)_

src/utils/graphQLClient.tsx
```diff
@@ -45,16 +102,80 @@ export const useAccessToken = () => {
...
+  return true;
+};
+
+const attemptTokenRefresh = async () => {
+  if (!currentAuth) {
+    throw new Error('Token refresh failed (401)');
+  }
+  const abortController = new AbortController();
+  refreshAbortController = abortController;
+  const timeoutId = globalThis.setTimeout(() => abortController.abort(), REFRESH_TIMEOUT_MS);
```
valentinpalkovic: **[Low] Style: `globalThis.setTimeout` here, `window.crypto` / `window.location` / `window.opener` elsewhere.**

For a browser-only Storybook addon, pick one. The rest of the codebase uses `window`, so prefer `window.setTimeout` / `window.clearTimeout` here for consistency.

_(Gemini + Claude)_

src/utils/graphQLClient.tsx
```diff
@@ -1,31 +1,88 @@
...
+import { type AuthStorage, AuthStorageSchema, refreshAccessToken } from './requestAccessToken';
 
+const REFRESH_TIMEOUT_MS = 10_000;
+const getStorage = () => (typeof localStorage === 'undefined' ? null : localStorage);
+
+let currentAuth: AuthStorage | null = null;
 let currentToken: string | null;
-let currentTokenExpiration: number | null;
-const setCurrentToken = (token: string | null) => {
+let refreshPromise: Promise<void> | null = null;
```
valentinpalkovic, line 21: **[Medium] Cross-tab refresh is not single-flight.**

`refreshPromise` is module-local; refresh tokens are shared across tabs via `localStorage`. Two Storybook tabs that 401 simultaneously will each attempt to refresh. If the backend rotates refresh tokens, the second tab's refresh will fail and log that tab out.

Options: coordinate via `BroadcastChannel`, listen for `storage` events to pick up newly-written `AuthStorage`, or document that this is acceptable if the backend doesn't rotate. Probably tolerable as a follow-up but worth a note.

_(Codex + Claude)_

src/utils/requestAccessToken.ts
```diff
@@ -9,10 +11,29 @@ export type TokenExchangeParameters = {
...
+  accessToken: z.string().min(1),
+  refreshToken: z.string().min(1),
+  tokenEndpoint: z.string().url(),
+  sessionId: z.string().min(1),
+});
+export type AuthStorage = z.infer<typeof AuthStorageSchema>;
+
+const TokenResponseSchema = z.object({
+  access_token: z.string().min(1),
+  refresh_token: z.string().min(1),
```
valentinpalkovic: **[Medium] `refresh_token` is required, but RFC 6749 §6 allows the server to omit it on refresh responses.**

```ts
const TokenResponseSchema = z.object({
  access_token: z.string().min(1),
  refresh_token: z.string().min(1),  // ← required
  ...
});
```

If Chromatic's token endpoint ever returns an access-only response (i.e. doesn't rotate refresh tokens on every refresh), the schema fails and the user is logged out. Make `refresh_token` optional and reuse the previous `refreshToken` when omitted:

```ts
const TokenResponseSchema = z.object({
  access_token: z.string().min(1),
  refresh_token: z.string().min(1).optional(),
  error: z.string().optional(),
  error_description: z.string().optional(),
});
```

Then in `refreshAccessToken`: `refreshToken: data.refresh_token ?? refreshToken`.

_(Codex + Claude)_

src/utils/requestAccessToken.ts
```diff
@@ -9,10 +11,29 @@ export type TokenExchangeParameters = {
...
   tokenEndpoint: string;
 };
 
+export const AuthStorageSchema = z.object({
+  version: z.literal(1),
+  accessToken: z.string().min(1),
+  refreshToken: z.string().min(1),
+  tokenEndpoint: z.string().url(),
+  sessionId: z.string().min(1),
+});
```
valentinpalkovic, line 26: **[Medium] No backward-compat for users with the previous raw-token storage format.**

The schema requires `version: literal(1)` and a JSON object. Existing users with a plain JWT string in `ACCESS_TOKEN_KEY` will hit the parse failure path → `clearCurrentAuth()` → forced re-login on upgrade.

If re-login is acceptable, please call it out in CHANGELOG and release notes (one-time hit but visible to users). Otherwise add a small migration — detect non-JSON value, treat as expired, drop it cleanly.

_(Codex)_

src/screens/Authentication/Verify.tsx
```diff
@@ -85,14 +85,15 @@ export const Verify = ({
             clientId,
             codeVerifier,
             redirectUri,
+            sessionId,
             tokenEndpoint,
             code: event.code,
           });
-          if (!token) throw new Error('Failed to fetch an access token');
-          accessToken.current = token;
+          setAuthenticatedSession(token);
```
valentinpalkovic: **[High] Refresh token persisted before the auth flow is actually complete.**

`setAuthenticatedSession(token)` here writes the refresh token to `localStorage` immediately after token exchange, but the comment three lines down still says "don't store it yet." For users with no projects, control branches into the `newProjectUrl` dialog (line ~113) without ever calling `setAccessToken`. If the user closes the dialog at that point, a long-lived refresh token sits in `localStorage` while the Panel-level auth state is still empty.

Fix: defer `setAuthenticatedSession` to the same point as `setAccessToken(accessToken.current)` (lines 104 / 124). Or, alternatively, clear auth on Verify-screen unmount unless `setAccessToken` was called.

_(Codex + Claude)_

eslint.config.mjs
```diff
@@ -71,6 +71,31 @@ export default [
...
+        },
+        {
+          selector: "CallExpression[callee.name='eval']",
+          message: 'eval is banned',
+        },
+        {
+          selector: "NewExpression[callee.name='Function']",
+          message: 'Function constructor is banned',
+        },
+      ],
```
valentinpalkovic, line 122: **[Low] Sink ban is incomplete as a guardrail.**

Great addition overall. Common sinks not yet covered:

- `outerHTML` assignment
- `document.write` / `document.writeln`
- iframe `srcdoc` JSXAttribute
- `Range.createContextualFragment`
- string-form `setTimeout` / `setInterval`

Suggested additions:

```js
{ selector: "AssignmentExpression[left.property.name='outerHTML']", message: 'outerHTML assignment is banned' },
{ selector: "CallExpression[callee.object.name='document'][callee.property.name=/^writeln?$/]", message: 'document.write/writeln is banned' },
{ selector: "JSXAttribute[name.name='srcdoc']", message: 'iframe srcdoc is banned' },
{ selector: "CallExpression[callee.property.name='createContextualFragment']", message: 'createContextualFragment is banned' },
{ selector: "CallExpression[callee.name=/^setTimeout|setInterval$/][arguments.0.type='Literal']", message: 'String-form setTimeout/setInterval is banned' },
```

_(Codex)_

src/screens/GuidedTour/PulsatingEffect.tsx
```diff
@@ -41,5 +41,5 @@ export function PulsatingEffect({ targetSelector }: { targetSelector: string }):
     };
   }, [targetSelector]);
 
-  return <></>;
+  return <React.Fragment />;
```
valentinpalkovic: **[Low] `<React.Fragment />` is inconsistent with the rest of the repo.**

Elsewhere in the codebase (e.g. `VisualTests.tsx:323`) we use `<></>`. Suggest reverting this hunk to keep the style uniform.

_(Gemini)_

valentinpalkovic (approved):

PR #436, opened 2026-06-08
Update Codecov to fix GPG issue

codykaup:
Codecov started failing with a GPG verification issue. This was updated in the later versions of their Action. 

Issue for reference: [https://github.com/codecov/codecov-action/issues/1956](https://github.com/codecov/codecov-action/issues/1956)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,6 +23,6 @@ jobs:
         run: yarn test
 
       - name: Upload coverage reports to Codecov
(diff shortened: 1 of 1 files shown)


thafryer:
@codykaup Is this still needed?

codykaup:
> @codykaup Is this still needed?

It seems the tests are still working so perhaps we don't. Not sure why this repo isn't hitting the same errors we hit on all other repos. 🤔 

We can close it for now and reopen if we start seeing issues.

PR #437, opened 2026-06-11
Add view telemetry for top-of-funnel share screens

shilman:
## What

The share flow reported telemetry from the `uploading` step onward, but the top of the funnel was dark: the welcome, sign-in, and SSO subdomain screens emitted nothing, so we couldn't measure drop-off before publish/auth.

This adds three view events, emitted via the existing `emitTelemetry` channel in `useShareExecution`:

| Screen | Telemetry before | Telemetry after |
|---|---|---|
| `welcome` | ❌ none | ✅ `share-welcome-viewed` |
| `idle` (sign-in) | ❌ none | ✅ `share-signin-viewed` |
| `subdomain` (SAML SSO) | ❌ none | ✅ `share-sso-viewed` |
| `uploading` | ✅ `share-initiated`, `share-url-received`, `share-url-copied`, `share-canceled` | unchanged |
| `complete` | ✅ `share-upload-completed`, `share-url-copied` | unchanged |
| `error` | ✅ `share-failed`, `share-auth-retry`, `share-auth-completed` | unchanged |

Semantics:
- One event per screen **entry** — a ref suppresses re-renders of the same screen, and leaving/returning (e.g. subdomain → back → idle) fires again.
- Skipped when a token is present, since welcome/idle/subdomain auto-skip straight to uploading and are never actually shown.
- Payload follows the existing `share-*` shape (`{ action, entryPoint: 'toolbar' }`).

## Manual QA

Ran the addon's own Storybook (`yarn storybook`, addon loaded from source) signed out, with a listener on the `chromaui/addon-visual-tests/telemetry` channel event, and walked: open Share popover → Publish & copy link → Connect with SAML SSO → Back. Captured event sequence (each exactly once per view):

```
share-welcome-viewed   (popover opened on welcome)
share-signin-viewed    (clicked "Publish & copy link" while signed out)
share-sso-viewed       (clicked "Connect with SAML SSO")
share-signin-viewed    (clicked "Back" — re-entry fires again)
```

All four also arrived at the Node-side `TELEMETRY` preset handler (logged as `eventType: addon-visual-tests` with `addonVersion` enrichment via `STORYBOOK_TELEMETRY_DEBUG`). Re-rendering the open popover produced no duplicate events. The post-auth events (`share-initiated` etc.) are not reachable in this signed-out QA pass; their code paths are untouched and remain covered by the existing unit tests.

## Tests

- 5 new tests in `SharePopup.test.tsx`: fires on entry, no double-fire on re-render, once per entry with re-fire on return, suppressed when signed in, and no view events for non-funnel screens.
- Test harness now persists `useRef` values across invocations (keyed by hook call order) so repeated calls behave like re-renders; existing 12 tests unchanged and passing.
- `yarn typecheck`, `yarn lint`, `yarn vitest run` (21 files, 176 tests) all green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/screens/SharePopup/SharePopup.test.tsx
+++ b/src/screens/SharePopup/SharePopup.test.tsx
@@ -19,12 +19,22 @@ let reducerState: ShareReducerState = {
   shareTriggeredId: null,
 };
 let shareProgressValue: ShareProgress | undefined = undefined;
+let tokenValue: string | null = 'token-123';
+
+// Refs persist across invocations (keyed by hook call order) so that calling
+// the component repeatedly behaves like re-rendering the same mount.
+const refStore: { current: unknown }[] = [];
+let refIndex = 0;
 
 vi.mock('react', async (importOriginal) => {
   const actual = await importOriginal<typeof import('react')>();
   return {
     ...actual,
-    useRef: (initial: unknown) => ({ current: initial }),
+    useRef: (initial: unknown) => {
+      const ref = (refStore[refIndex] ??= { current: initial });
+      refIndex += 1;
+      return ref;
+    },
     useCallback: (fn: unknown) => fn,
     // useEffect: run synchronously so effects fire during the "render" call
     useEffect: (fn: () => (() => void) | void) => {
@@ -41,7 +51,7 @@ vi.mock('storybook/manager-api', () => ({
 }));
 
 vi.mock('../../utils/graphQLClient', () => ({
-  useAccessToken: () => ['token-123', mocks.updateToken],
+  useAccessToken: () => [tokenValue, mocks.updateToken],
 }));
 
 vi.mock('../../auth/authStore', () => ({
@@ -88,6 +98,7 @@ function makeApi() {
 
 // Call the component as a plain function to trigger all useEffect calls
 function invokeSharePopup() {
+  refIndex = 0;
   (SharePopup as any)({ api: makeApi() });
 }
 
@@ -108,7 +119,13 @@ afterEach(() => {
     shareTriggeredId: null,
   };
   shareProgressValue = undefined;
+  tokenValue = 'token-123';
+  refStore.length = 0;
 });
+
+function viewEmits(action: string) {
+  return mocks.channel.emit.mock.calls.filter(([, payload]) => payload?.action === action);
+}
 
 describe('SharePopup', () => {
   describe('stale shareProgress filtering by shareRequestId', () => {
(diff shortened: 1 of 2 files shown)


shilman:
## Canary verification ✅

**Canary:** `@chromatic-com/storybook@5.2.2--canary.9c6dc66.0`

Provenance: published by this repo's [Release workflow run](https://github.com/chromaui/addon-visual-tests/actions/runs/27359549896) on this branch via `auto shipit`. `npm view @chromatic-com/storybook@5.2.2--canary.9c6dc66.0 gitHead` returns `9c6dc663446e2e19665caf7d5f653a46962315cd` — this PR's head commit — and the published `dist/manager.mjs` contains the new event names.

Installed the canary from npm into a fresh `react-ts` Vite project (`npm create vite` + `npx storybook init`) and walked the share flow signed out, with a listener on the `chromaui/addon-visual-tests/telemetry` channel event:

| Step | Event captured |
|---|---|
| Open Share popover (welcome) | `share-welcome-viewed` |
| Close + reopen popover | `share-welcome-viewed` (fires again per view, no duplicates while open) |
| "Publish & copy link" (signed out → sign-in) | `share-signin-viewed` |
| "Connect with SAML SSO" | `share-sso-viewed` |
| "Back" (return to sign-in) | `share-signin-viewed` |

Each event fired exactly once per screen view. End-to-end flow confirmed against the published package, not just the source tree.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

shilman:
**AI-generated**

@ghengeveld flagging you here for a stricter maintainability read.

I think the behavior is correct, but I would not approve this structure as-is under a code-quality bar.

1. `useShareExecution` is becoming a kitchen-sink side-effect coordinator. This PR adds screen-view telemetry at `src/screens/SharePopup/useShareExecution.ts:105`, but the hook already owns auto-skip, start-share emission, auth retry, progress effects, and publish/cancel handlers. The cleaner move is to put screen-entry telemetry at the screen/render boundary, either as a small `useShareViewTelemetry(...)` hook beside `SharePopup` or a keyed wrapper around the `SharePopup.tsx` switch. Then “screen was entered” is modeled by the UI state transition instead of another manual ref inside execution flow.

2. The test change adds a hand-rolled React `useRef` runtime keyed by hook call order at `src/screens/SharePopup/SharePopup.test.tsx:24`. That is too magical for a test whose contract is specifically about re-render and screen-entry semantics. Reordering or adding refs in `useShareExecution` can silently bind old values to the wrong ref, and `afterEach` clears `refStore` without resetting `refIndex`. I would rather extract the screen-view telemetry logic into a tiny pure helper/hook and test that directly, or use a real React render/rerender harness instead of expanding this fake renderer.

No file-size threshold issue here, and the focused tests pass locally:

```bash
npx vitest run src/screens/SharePopup/SharePopup.test.tsx src/screens/SharePopup/shareMachine.test.ts
```

But I think these two structural issues are worth fixing before approval.

ghengeveld (approved):

PR #438, opened 2026-07-06
Render DPR 2 snapshots at the correct display size

ghengeveld:
## Summary

DPR 2 captures store images at 2× physical pixel dimensions. VTA was rendering them using raw `imageWidth` / `imageHeight`, so high-DPI snapshots appeared twice as wide as they should.

This mirrors the Chromatic webapp approach: fetch `deviceScaleFactor` from the public API and scale display dimensions via `getDisplayImageSize`. Includes a `HighDPR` `SnapshotImage` story with a real DPR 2 capture fixture (2400×1246 → 1200×623 at DPR 2).

Depends on https://github.com/chromaui/chromatic/pull/11500.

## How to QA

1. Run `yarn vitest run src/utils/dpr.test.ts`.
2. Open Storybook → **Webapp components / SnapshotImage → HighDPR** and confirm the snapshot renders at logical size (not 2× too wide).
3. With a DPR 2–enabled project linked in VTA, open snapshot comparison and confirm images match Chromatic webapp sizing.

Check ☑️ the appropriate box once you've finished QA:

- [x] Author - QA OK
- [ ] Reviewer - QA OK


src/utils/dpr.ts
```diff
@@ -0,0 +1,15 @@
+export const getDisplayImageSize = ({
+  captureImageSize,
+  deviceScaleFactor = 1,
+}: {
+  captureImageSize?: { width: number; height: number };
+  deviceScaleFactor?: number;
+}): { width: number; height: number } | undefined => {
+  if (!captureImageSize || deviceScaleFactor <= 1) {
```
treeforever, line 8: This looks good for web! One general question, does VTA have possibilities of rendering mobile snapshots at all? The current implementation in webapp covers the mobiel snapshots [here](https://github.com/chromaui/chromatic/blob/main/services/webapp/lib/dpr.ts#L15). But I am not sure what's VTA's mobile prospect. Not a blocker, more a general question for the future direction

treeforever (approved):
The code looks good to me. I left a general direction question about covering mobile snapshots or not.
QA I am not sure how to get the VTA work here. I have a production app with DPR 2 enabled. Let me know if that's a [project](https://www.chromatic.com/builds?appId=6a32c6a3432d033dad033a61) you look for?
One thing I noticed in VTA testing locally is that it's not easy to spot the display size difference when the snapshot is page size because it's capped by a max width. Small component like a button or a badge is more obviously to see that the snapshot is twice as big. Just as a side note as I see the added story is with a full page snapshot.

src/utils/dpr.ts
ghengeveld: It's a fair point. I deliberately omitted the captured pixel ratio logic because I considered RN/iOS/Android out of scope, but seemingly we do technically support React Native, so perhaps we need to revise that. I'll check with @shilman.

shilman: No. AFAIK:
- A single RN project can contain an RN Storybook, a RNW Storybook, or both.
- Addons like VTA only run in Web storybooks, not in SBRN (technically SBRN has its own janky addon system, but it's very limited and may as well not exist for the sake of this discussion).
- Potentially the same project can contain a Chromatic RN project, a Chromatic Web project, or both.

The only (super edge case) scenario worth considering when a project contains both RN and RNW Storybooks and only a Chromatic RN project. I suppose the user could configure VTA to show snapshots from the RN project in that case. But I don't think anybody would actually set things up this way, unless it was by accident.

ghengeveld: Great, let's just omit that extra complexity then. We can always add it later if it turns out to be necessary.

ghengeveld:
> The code looks good to me. I left a general direction question about covering mobile snapshots or not. QA I am not sure how to get the VTA work here. I have a production app with DPR 2 enabled. Let me know if that's a [project](https://www.chromatic.com/builds?appId=6a32c6a3432d033dad033a61) you look for?

I used [this project here](https://www.chromatic.com/library?appId=68fa77736388f611613cf3a2&buildNumber=17) to find a DPR 2.0 snapshot. We cannot QA this until the API change rolls out, unfortunately.

> One thing I noticed in VTA testing locally is that it's not easy to spot the display size difference when the snapshot is page size because it's capped by a max width. Small component like a button or a badge is more obviously to see that the snapshot is twice as big. Just as a side note as I see the added story is with a full page snapshot.

You can change the Theme to light or dark to get it to render at full screen width. By default it renders in a narrow "addon panel" viewport.

src/utils/dpr.ts
treeforever: awesome, thank you for clarifying that! @ghengeveld @shilman !

treeforever:
@ghengeveld thanks for your replies. Your other [PR](https://github.com/chromaui/chromatic/pull/11500) in webapp has been QAed and merged by me. Please feel free to proceed with this PR. Thank you!

